PR #1, opened 2019-11-06
Expand AST and add basic well-formed checking

LucasPickering:
Wrote some dank shit. 

Set it up so you can define an AST in JSON and "compile" it. It just does the checking then spits the AST back out. Try this out:

`input.json`
```json
{
  "instructions": [
    { "Set": { "Int": 5 } },
    { "Create": "test" },
    { "Push": "test" },
    { "Pop": "test" }
  ]
}
```

`cargo run -- compile --input input.json`

It should print out basically the same thing. Neat!

### DIFF
--- a/lib/src/compiler/well_formed.rs
+++ b/lib/src/compiler/well_formed.rs
@@ -1,16 +1,93 @@
 use crate::{
-    ast::{Ast, WellFormedAst},
+    ast::{
+        RawBody, RawInstruction, RawProgram, WellFormedBody,
+        WellFormedInstruction, WellFormedProgram,
+    },
     compiler::Compiler,
+    env::Env,
 };
 use failure::Error;
 
-impl Compiler<Ast> {
+fn check_body_well_formed(
+    body: RawBody,
+    env: &mut Env,
+) -> Result<WellFormedBody, Error> {
+    let mut result = Vec::new();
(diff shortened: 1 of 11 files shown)


lib/src/compiler/well_formed.rs
```diff
@@ -1,16 +1,93 @@
...
+
+    #[test]
+    fn test_no_errors() {
+        let compiler = Compiler(RawProgram {
+            body: vec![
+                RawInstruction::Create("id".into()),
+                RawInstruction::Set(RawValue::Int(5)),
+                RawInstruction::Push("id".into()),
+                RawInstruction::Pop("id".into()),
+                RawInstruction::Destroy("id".into()),
```
JRMurr: Whats all this `into` stuff for?

LucasPickering: Defined here: https://doc.rust-lang.org/std/string/struct.String.html#impl-From%3C%26%27_%20str%3E
`From` is the inverse of `Into`: generally you implement from, then it automatically implements `Into` for you. In this case, it converts `&str` into `String`

JRMurr:
@LucasPickering confiicts

LucasPickering:
@JRMurr fixed, you bitch

PR #3, opened 2019-11-21
Add basic execution layout for the game

LucasPickering:
Added a framework for step-by-step execution of game programs. Here's the TL;DR:

- `Program` and `Environment` are the input types
- You create a `Machine` from a program and env
- You can execute a machine step-by-step with `execute_next`

You can test this out by running this program:

```
{
  "body": [
    "Read",
    {
      "Set": 2
    },
    "Write",
    "Read",
    {
      "Set": 3
    },
    "Write",
    "Read",
    {
      "Set": 4
    },
    "Write"
  ]
}
```

Save as `prog.json`, then run `cargo run -- -i prog.json`. Should print some #litty shit.

It might also be useful to look through the generated docs. Pull down the branch and run `cargo doc --open`


lib/src/machine.rs
```diff
@@ -0,0 +1,379 @@
...
+/// All stack manipulation logic should be implemented here, to keep it
+/// contained and scalable.
+/// This is zero-cost at runtime, see [newtype pattern](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html).
+#[derive(Debug, PartialEq)]
+pub struct Stacks(Vec<Vec<LangValue>>);
+
+impl Stacks {
+    fn new(env: &Environment) -> Self {
+        Self(
+            iter::repeat(0)
```
JRMurr: why the repeat(0)?

lib/src/machine.rs
```diff
@@ -0,0 +1,379 @@
...
+        env: &Environment,
+        stack_id: StackIdentifier,
+        value: LangValue,
+    ) -> Fallible<()> {
+        let stack = self.get_stack(stack_id)?;
+
+        // If the stack is capacity, make sure we're not over it
+        if let Some(max_stack_size) = env.max_stack_size {
+            if stack.len() == max_stack_size {
+                return Err(RuntimeError::StackOverflow(stack_id).into());
```
JRMurr: motion to make the error display this image https://i.imgur.com/Jmjv5To.jpg

lib/src/machine.rs
```diff
@@ -0,0 +1,379 @@
...
+        }
+    }
+}
+
+/// A recursive structure to track the progression of a program. A program
+/// counter consist of a flat list of instructions that need to be run, plus
+/// an optional child `ProgramCounter`. If the child is present, the child's
+/// instructions will always be executed before any of the parent's. Once all
+/// of a program counter's instructions have been executed, it's considered
+/// exhausted.
```
JRMurr: Very clean way to handle blocks. I approve with 4.9 belching burps out of 5

lib/src/machine.rs
LucasPickering: What I'm trying to do here is a create a `Vec` of `Vec`s, where the length of the parent vec is `env.num_stacks` (and each sub-vec is one of the user's stacks). `iter::repeat` repeats the given value infinitely, so together `iter::repeat(0).take(env.num_stacks)` creates an iter that repeats `0` by `env.num_stacks` times. The `0` isn't important, I just picked a value. Once it has that, it generates one new vec per element. As I'm typing this though, I realized there's a slightly cleaner way to do this, so I'm going to update it and add a comment

lib/src/machine.rs
LucasPickering: oh fuqqqqq ye bud

lib/src/machine.rs
LucasPickering: aww hell yeah, we can do that on the frontend

PR #5, opened 2019-11-27
Add websocket server

LucasPickering:
- Changed up the error types in lib. I made them more restrictive so instead of just using `Fallible` everywhere, we specifically use our own `CompileError` and `RuntimeError` types. This makes it possible to serialize the errors so they can be sent over the websocket. I actually don't think we need `failure` any more now, but I'm leaving it in for now. Will prob remove it in another PR.
- Added a basic websocket server
- Added a shitty script for running programs via the server. 

Here's an example program you can test:
```json
{
  "body": [
    "Read",
    "Write",
    "Read",
    "Write",
    "Read",
    "Write"
  ]
}
```

copy it, then do:

```sh
cd server
# in one window
cargo run
# in another window
paste > prog.json # or however you wanna make the file
./test_socket.py -p ./prog.json # might need to install the websockets pkg
```

This should spit out some output like this:

```
<<< {"type": "edit", "content": {"source": "{\n  \"body\": [\n    \"Read\",\n    \"Write\",\n    \"Read\",\n    \"Write\",\n    \"Read\",\n    \"Write\"\n  ]\n}"}}
>>> {'type': 'source', 'content': {'source': '{\n  "body": [\n    "Read",\n    "Write",\n    "Read",\n    "Write",\n    "Read",\n    "Write"\n  ]\n}'}}
<<< {"type": "compile", "content": null}
>>> {'type': 'machine_state', 'content': {'state': {'input': [3, 2, 1], 'output': [], 'workspace': 0, 'stacks': []}, 'is_complete': False, 'is_successful': False}}
<<< {"type": "step", "content": null}
>>> {'type': 'machine_state', 'content': {'state': {'input': [3, 2], 'output': [], 'workspace': 1, 'stacks': []}, 'is_complete': False, 'is_successful': False}}
<<< {"type": "step", "content": null}
>>> {'type': 'machine_state', 'content': {'state': {'input': [3, 2], 'output': [1], 'workspace': 1, 'stacks': []}, 'is_complete': False, 'is_successful': False}}
<<< {"type": "step", "content": null}
>>> {'type': 'machine_state', 'content': {'state': {'input': [3], 'output': [1], 'workspace': 2, 'stacks': []}, 'is_complete': False, 'is_successful': False}}
<<< {"type": "step", "content": null}
>>> {'type': 'machine_state', 'content': {'state': {'input': [3], 'output': [1, 2], 'workspace': 2, 'stacks': []}, 'is_complete': False, 'is_successful': False}}
<<< {"type": "step", "content": null}
>>> {'type': 'machine_state', 'content': {'state': {'input': [], 'output': [1, 2], 'workspace': 3, 'stacks': []}, 'is_complete': False, 'is_successful': False}}
<<< {"type": "step", "content": null}
>>> {'type': 'machine_state', 'content': {'state': {'input': [], 'output': [1, 2, 3], 'workspace': 3, 'stacks': []}, 'is_complete': True, 'is_successful': True}}
```

Shit's #lit dude


server/src/websocket.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    /// Processes the given text message, and returns the appropriate response
+    /// event. The return type on this is a little funky because all our
+    /// event types (OK and error) are under the same enum. We still use a
+    /// Result because it makes it easier to exit early in the case of an error.
+    fn process_msg(
+        &mut self,
+        text: String,
+    ) -> Result<OutgoingEvent, OutgoingEvent> {
```
JRMurr: im still confused on why its needs to be here twice

LucasPickering: It doesn't _need_ to be, but by returning a result, it makes it easier to handle errors. There are 3 spots in the function where we use the `?` operator, which can only be used when you're returning a `Result` (mostly). Conceptually from the outside it would make more sense to just return one `OutgoingEvent`, but if we did that, then the error-handling cases would have to look like
```rust
let socket_msg = match serde_json::from_str::<IncomingEvent>(&text) {
    Ok(msg) => msg,
    Err(err) => return err.into(),
};
```

and I didn't really want to write that in 3 different places.

JRMurr: Im still a lil confused but i trust you lol

PR #6, opened 2019-11-28
Add basic parsing

JRMurr:
Initial parsing pass through. Errors are pretty ass atm but i think i know how to make them better in a follow up

I put some tests with example programs. Right now i only allow the uppercase Read,Write, etc but it'll be easy to switch to make it lowercase or allow mixed-cased since that's the only thing they can put in.

Also right now it does not support the program having comments

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -12,3 +12,4 @@ failure = "0.1"
 itertools = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+nom = "5.0"
--- a/lib/src/compiler/parse.rs
+++ b/lib/src/compiler/parse.rs
@@ -1,5 +1,106 @@
-use crate::{ast::Program, compiler::Compiler, error::CompileError};
+use crate::{
+    ast::{Instr, LangValue, Program, StackIdentifier},
+    compiler::Compiler,
+    error::CompileError,
+};
+use nom::{
+    branch::alt,
+    bytes::complete::tag_no_case,
+    character::complete::{char, digit1, multispace0},
+    combinator::map_res,
+    multi::{many0, many1},
+    sequence::{delimited, preceded, terminated},
+    IResult,
+};
 use std::io::Read;
+
+fn parse_read(input: &str) -> IResult<&str, Instr> {
+    // tag_no_case returns a (str, str) tuple
+    // first element is whats left to parse
+    // second is what matched
+    let (input, _) = tag_no_case("Read")(input)?;
+    Ok((input, Instr::Read))
+}
+
+fn parse_write(input: &str) -> IResult<&str, Instr> {
+    let (input, _) = tag_no_case("Write")(input)?;
+    Ok((input, Instr::Write))
+}
+
+fn parse_set(input: &str) -> IResult<&str, Instr> {
+    let (input, _) = tag_no_case("Set")(input)?;
+    let (input, val) = preceded(
+        multispace0,
+        map_res(digit1, |s: &str| s.parse::<LangValue>()),
+    )(input)?;
+    Ok((input, Instr::Set(val)))
+}
+
+fn parse_push(input: &str) -> IResult<&str, Instr> {
+    let (input, _) = tag_no_case("Push")(input)?;
+    let (input, val) = preceded(
+        multispace0,
+        map_res(digit1, |s: &str| s.parse::<StackIdentifier>()),
+    )(input)?;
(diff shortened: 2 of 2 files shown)


lib/Cargo.toml
```diff
@@ -12,3 +12,4 @@ failure = "0.1"
 itertools = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
+nom = "5.0.1"
```
LucasPickering: I think we should just use `major.minor`, so that we don't have to make changes to pick up patch versions.

lib/src/compiler/parse.rs
```diff
@@ -1,14 +1,220 @@
...
+    character::complete::{char, digit1, multispace0},
+    combinator::map_res,
+    multi::{many0, many1},
+    sequence::{delimited, preceded, terminated},
+    IResult,
+};
 use std::io::Read;
 
+fn parse_read(input: &str) -> IResult<&str, Instr> {
+    let (input, _) = tag("Read")(input)?;
```
LucasPickering: What is the `_` value that you're throwing away in every function?

lib/src/compiler/parse.rs
```diff
@@ -1,14 +1,220 @@
...
+    let (input, _) = tag("Write")(input)?;
+    Ok((input, Instr::Write))
+}
+
+fn parse_set(input: &str) -> IResult<&str, Instr> {
+    let (input, _) = tag("Set")(input)?;
+    let (input, val) = preceded(
+        multispace0,
+        map_res(digit1, |s: &str| s.parse::<LangValue>()),
+    )(input)?;
```
LucasPickering: Can you add a comment on each one of these with a little example of what it's parsing? idk what all this `preceded` and `multispace0` mumbo jumbo is.

lib/src/compiler/parse.rs
```diff
@@ -1,14 +1,220 @@
...
+fn parse_body(input: &str) -> IResult<&str, Vec<Instr>> {
+    let (input, res) = delimited(
+        preceded(multispace0, char('{')),
+        many0(try_each),
+        preceded(multispace0, char('}')),
+    )(input)?;
+    Ok((input, res))
+}
+
+fn my_parse(input: &str) -> IResult<&str, Vec<Instr>> {
```
LucasPickering: Can you rename this to be a bit more descriptive?

lib/src/error.rs
```diff
@@ -19,3 +19,9 @@ pub enum RuntimeError {
     #[fail(display = "Stack {} is empty", 0)]
     EmptyStack(StackIdentifier),
 }
+
+#[derive(Debug, Fail)]
+pub enum ParseError {
+    #[fail(display = "Parse error: {}", 0)]
+    Generic(String),
+}
```
LucasPickering: Just a pro tip, whoever merges their PR second is gonna have to change this, so merge this with the `CompileError` that I made. We'll also have to change `Compiler::parse` to return a `Result<..., CompileError>` instead of a `Fallible`

lib/src/compiler/parse.rs
```diff
@@ -1,14 +1,220 @@
...
+    }
+
+    #[test]
+    fn test_parse_while() {
+        assert_eq!(
+            my_parse(
+                "while {
+            Read
+            Write
+        }"
```
LucasPickering: I'd be down to support either any casing, so you could write `WHILE {...}` or `while {...}` or whatever, or just all caps, since that feels old and memey.

lib/src/compiler/parse.rs
JRMurr: got u fam

lib/Cargo.toml
JRMurr: o littty

lib/src/compiler/parse.rs
JRMurr: most of the builtin funcs return a touple of (str,str). The first element is whats left to parse, the second is the string of what matched. So i just through away the match (since it was Read) in this case then return the actual instruction. Ill add a comment

lib/src/compiler/parse.rs
JRMurr: On it

lib/src/compiler/parse.rs
JRMurr: making it not care about case. Maybe we should require cases like `MeMeS`

LucasPickering: spongebob case? that'd be dank lul

PR #8, opened 2019-12-01
Add postgres setup

LucasPickering:
- Added the postgres DB
- Added in the ORM to manage stuff
- Made the websocket read the environment from the DB


api/src/lang/machine.rs
```diff
@@ -122,8 +133,8 @@ impl MachineState {
 #[derive(Debug)]
 pub struct Machine {
     // Static data
-    env: Environment,
     program: Vec<MachineInstr>,
+    expected_output: Vec<LangValue>,
```
LucasPickering: I removed the `env` from here, because I was running into more lifetime issues when a `ProgramWebsocket` tried to hold a `Machine` and the `Environment` itself.

api/src/models/env.rs
```diff
@@ -1,14 +1,21 @@
...
+    pub id: i32,
+
+    // These two need to be i32s because postgres has no unsigned type.
+    // The insertion code and DB should both enforce that they are >= 0.
     /// Maximum number of stacks permitted
-    pub num_stacks: usize,
+    pub num_stacks: i32,
     /// Maximum size of each stack. If None, the capacity is unlimited.
-    pub max_stack_size: Option<usize>,
+    pub max_stack_size: Option<i32>,
```
LucasPickering: Had to change these to `i32`s, because postgres has no unsigned type so `usize` doesn't can't map to anything in postgres.

PR #16, opened 2019-12-06
Add registers and improve AST

LucasPickering:
- Combined some common parts of the two ASTs by pulling most of the instructions out into a common `Operator` type
- Added registers and used them for all the instructions
- lit

### DIFF
--- a/api/src/lang/machine.rs
+++ b/api/src/lang/machine.rs
@@ -1,7 +1,7 @@
 use crate::{
     debug,
     error::RuntimeError,
-    lang::ast::{LangValue, MachineInstr, StackIdentifier},
+    lang::ast::{LangValue, MachineInstr, Operator, Register, StackIdentifier},
     models::Environment,
 };
 use serde::Serialize;
@@ -105,8 +105,8 @@ pub struct MachineState {
     /// The current output buffer. This can be pushed into, but never popped
     /// out of.
     pub output: Vec<LangValue>,
-    /// The mutable variable that the programmer has direct control over.
-    pub workspace: LangValue,
+    /// TODO
+    pub registers: Vec<LangValue>,
     /// The series of stacks that act as the programs RAM. The number of stacks
     /// and their capacity is determined by the initialization environment.
     pub stacks: Stacks,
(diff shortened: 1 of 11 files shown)


api/src/lang/machine.rs
```diff
@@ -117,10 +117,32 @@ impl MachineState {
...
         input.reverse(); // Reverse the vec so we can pop off the values in order
         Self {
-            stacks: Stacks::new(&env),
             input,
             output: Vec::new(),
-            workspace: 0,
+            registers: iter::repeat(0)
+                .take(usize::try_from(env.num_registers).unwrap())
+                .collect(),
+            stacks: Stacks::new(&env),
```
JRMurr: still confused on this iter.repeat(0) stuff

api/src/lang/parse.rs
```diff
@@ -1,101 +1,154 @@
...
 };
 
-fn parse_read(input: &str) -> IResult<&str, Instr> {
-    // tag_no_case returns a (str, str) tuple
-    // first element is whats left to parse
-    // second is what matched
-    let (input, _) = tag_no_case("Read")(input)?;
-    Ok((input, Instr::Read))
+fn arg_delim(input: &str) -> IResult<&str, char> {
+    one_of(" \t")(input)
```
JRMurr: I dont think this will catch multiple spaces, if we want them to be able to do that

LucasPickering: Yeah so I wanted to make it match multiple spaces without matching newlines and just didn't bother figuring out the repeating part, I figured I'd leave that to you. Are you cool with leaving it for now and you can change it in your PR?

api/src/lang/machine.rs
LucasPickering: Repeat just infinitely repeats whatever value you give it, so in this case it's repeating 0 once per register.

JRMurr: o that makes more sense

api/src/lang/parse.rs
JRMurr: yea thats fine

PR #18, opened 2019-12-09
Make parse errors nicer

JRMurr:
Started trying to make errors a little nicer. Still needs more work but since it changed up a lot i figured i'd push now.

I added `cut` after parsing tags that have arguments so we fail out of `alt` instead of trying to parse other options. Also added `context` that makes the errors have a custom name instead of `Tag` or `Alt`

### DIFF
--- a/api/src/lang/parse.rs
+++ b/api/src/lang/parse.rs
@@ -1,5 +1,5 @@
 use crate::{
-    error::{CompileError, CompileErrors},
+    error::CompileError,
     lang::{
         ast::{
             Instr, LangValue, Operator, Program, RegisterRef, StackIdentifier,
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -34,6 +34,11 @@ docker-compose up
 # Wait for DB to start up
 ./initdb.sh
 ```
+### TODO BETTER SEEDING
+```
+docker exec -it gdlk_db_1 psql gdlk
+insert into environments (num_stacks, max_stack_size, input, expected_output) values (0, 0, '{}', '{}');
+```
```
LucasPickering: Already adeed a script for this, you can take this out

api/src/lang/parse.rs
```diff
@@ -12,22 +12,30 @@ use crate::{
 use nom::{
     branch::alt,
     bytes::complete::tag_no_case,
-    character::complete::{char, digit1, multispace0, one_of},
-    combinator::{all_consuming, map, map_res},
+    character::complete::{
+        alpha1, char, digit1, multispace0, multispace1, one_of,
+    },
```
LucasPickering: unused imports

api/src/lang/parse.rs
```diff
@@ -88,123 +103,117 @@ fn parse_value_source(input: &str) -> IResult<&str, ValueSource> {
...
 
-fn parse_read(input: &str) -> IResult<&str, Instr> {
+fn parse_read(input: &str) -> IResult<&str, Instr, VerboseError<&str>> {
     // input is remaining stuff to parse
     // tuple is output values, we throw away the first two because that's
     // "Read" and the whitespace delim
     // >>> Read RX0
-    let (input, (_, _, reg)) =
-        tuple((instr("Read"), arg_delim, reg_ident))(input)?;
+    let (input, _) = context("Read", instr("Read"))(input)?;
```
LucasPickering: Why not put the `context` call inside `instr()` instead, so you don't have to repeat it for every instruction?

api/src/lang/parse.rs
```diff
@@ -88,123 +103,117 @@ fn parse_value_source(input: &str) -> IResult<&str, ValueSource> {
...
-            parse_add,
-            parse_sub,
-            parse_mul,
-            parse_push,
-            parse_pop,
-            parse_if,
-            parse_while,
-        )),
+        context(
+            "Expression",
```
LucasPickering: Can you change this to `"Instruction"`?

api/src/lang/parse.rs
JRMurr: Big brain

PR #20, opened 2019-12-12
Split Environment into two separate models

LucasPickering:
Split the `Environment` model into `HardwareSpec` and `ProgramSpec`. Open to suggestions on the names. I also improved the DB seeding system a little bit, so it's easier to update now. I have some ideas to make it more betterer, will work on that at some point. Also, I confirmed this is all still working with `test_socket.py`. I loaded the initial data with `./initdb.sh`, then ran `./test_socket.py -s prog.gdlk` with this program:

```
WHILE RLI {
    READ RX0
    WRITE RX0
}
```

and the shit ran and passed. Pog


api/src/server/websocket.rs
```diff
@@ -215,18 +221,29 @@ impl StreamHandler<ws::Message, ws::ProtocolError> for ProgramWebsocket {
...
+    // As far as I can tell, Diesel doesn't have an easier way to do a foreign
+    // key lookup
+    let hardware_spec: HardwareSpec = hardware_specs::dsl::hardware_specs
+        .find(program_spec.hardware_spec_id)
+        .get_result(conn)
+        .map_err(ServerError::from)?;
+    ws::start(
+        ProgramWebsocket::new(hardware_spec, program_spec),
+        &r,
+        stream,
```
JRMurr: Could you do a join so you can do it in one query?

JRMurr:
spec feels a little weird in program_spec how about hardware_env and program_env?

api/src/server/websocket.rs
LucasPickering: Yeah so normally I would've but the Diesel docs actually recommend using separate queries:
> Associated data is typically loaded in multiple queries (one query per table). This is usually more efficient than using a join, especially if 3 or more tables are involved. For most datasets, using a join to load in a single query transmits so much duplicate data that it costs more time than the extra round trip would have.
As I'm typing this tho, I'm realizing that might only refer to one-to-many relationships, not when you're just fetching a single related row. I'll look into doing a join, and if it's not too hard I'll do that.
https://docs.rs/diesel/1.4.3/diesel/associations/index.html

LucasPickering:
> 
> 
> spec feels a little weird in program_spec how about hardware_env and program_env?

Idk, I prefer spec, because it's a specification that the program has to meet. I feel like environment is a more ambiguous term. Maybe there's a different word we could use that would be best? I'll think about it.

PR #21, opened 2019-12-13
Split out core crate

LucasPickering:
Pulled non-web related stuff out into the core crate. Also split all the `mod.rs` tests out into a separate integration test directory (`core/tests` now). Shit's lit yo. Also improved the docs a bit. Confirmed that the webserver is still working too.


JRMurr:
pog

PR #22, opened 2019-12-16
Add CMP instruction, add insertion sort test, clean up a few other things

LucasPickering:
- Add the CMP instruction
- Add a test that runs insertion sort
- Some clean up in `MachineState`

### DIFF
--- a/core/src/parse.rs
+++ b/core/src/parse.rs
@@ -46,6 +46,25 @@ where
     tuple((one_arg(arg_parser_one), one_arg(arg_parser_two)))
 }
 
+fn three_args<I, O1, O2, O3, E: ParseError<I>, F, G, H>(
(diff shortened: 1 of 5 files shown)


core/src/parse.rs
```diff
@@ -111,9 +130,7 @@ fn lang_value(input: &str) -> IResult<&str, LangValue, VerboseError<&str>> {
 }
 
 /// Parses either a `LangValue` or `Register`.
-fn parse_value_source(
-    input: &str,
-) -> IResult<&str, ValueSource, VerboseError<&str>> {
+fn value_source(input: &str) -> IResult<&str, ValueSource, VerboseError<&str>> {
```
LucasPickering: tryna stick with the convention from the other utility parsers

PR #24, opened 2019-12-18
Add comment parsing and make sure instrs have new line

JRMurr:


core/src/parse.rs
```diff
@@ -257,11 +282,11 @@ fn parse_gdlk(input: &str) -> IResult<&str, Vec<Instr>, VerboseError<&str>> {
     // parses the whole program followed by 0 or more whitespace chars
 
     // consume starting whitespace
-    let (input, _) = multispace0(input)?;
+    let (input, _) = comment_or_spaces(input)?;
     // make sure something is there but don't consume the input
     // TODO: make this error message nicer
     peek(alpha1)(input)?;
-    let (input, res) = all_consuming(many0(try_each))(input)?;
+    let (input, res) = all_consuming(many0(try_each_instr))(input)?;
```
LucasPickering: I just noticed this function doesn't call `parse_body`. It prob should since a body is just a vec of instructions so the program as a whole is a body. Would that be possible/easy?

core/src/parse.rs
```diff
@@ -215,9 +216,33 @@ fn parse_while(input: &str) -> IResult<&str, Instr, VerboseError<&str>> {
...
+    Ok((input, ""))
+}
+
+fn comment_or_spaces(input: &str) -> IResult<&str, &str, VerboseError<&str>> {
+    let (input, _) =
+        many0(tuple((space0, many1(alt((comment, line_ending))), space0)))(
+            input,
+        )?;
+    Ok((input, ""))
+}
```
LucasPickering: Could we combine these two funcs?

core/src/parse.rs
```diff
@@ -215,9 +216,33 @@ fn parse_while(input: &str) -> IResult<&str, Instr, VerboseError<&str>> {
...
+// TODO: for now throwing away spaces and comments
+// probably want too keep them when we do source mapping
+fn comment(input: &str) -> IResult<&str, &str, VerboseError<&str>> {
+    let (input, _) = preceded(
+        space0,
+        context(
+            "Comment",
+            terminated(
+                char(';'),
+                cut(tuple((take_till(|c| c == '\n'), line_ending))),
```
LucasPickering: Why do you need both the`c == '\n'` and `line_ending`? If possible we should just use `line_ending`, for portability (I highly doubt this thing will ever have to process a `\r`, but just in case...

core/src/parse.rs
JRMurr: It could work, but i have logic in parse_body that looks for `{` and `}`. I feel like doing it this way makes the most sense imo

core/src/parse.rs
JRMurr: We could but i think this way makes it easier to see whats going on

core/src/parse.rs
JRMurr: I tried to use `not_line_ending` and `many0` but was getting a bunch of weird issues. Ill just add `\r` in that check too

core/src/parse.rs
LucasPickering: ah gotcha, that's fine then

core/src/parse.rs
```diff
@@ -125,9 +125,15 @@ fn stack_ident(
...
+/// whitespace.
 fn lang_value(input: &str) -> IResult<&str, LangValue, VerboseError<&str>> {
-    map_res(digit1, |s: &str| s.parse::<LangValue>())(input)
+    let (input, is_neg) = opt(char('-'))(input)?;
+    let (input, val) =
+        map_res(digit1, |s: &str| s.parse::<LangValue>())(input)?;
+    match is_neg {
+        Some(_) => Ok((input, -val)),
+        None => Ok((input, val)),
+    }
```
LucasPickering: I think there's a small bug here: I don't think we'll be able to parse the min value (`-2147483648`). Since it'll try to parse `2147483648` to an `i32`, which isn't valid. This is fine for now, but can you add tests that include `LangValue::MIN` and `LangValue::MAX`, and just comment out the `MIN` one so we remember to fix it?

JRMurr: got it to work for mins

PR #30, opened 2020-01-22
Actix web 2.0

LucasPickering:
Upgraded actix to 2.0 (async!). Also fixed some websocket stuff. Turns out that multiple route params get passed in as a tuple instead of separate params. The changes were pretty simple, just had to add async to some stuff. I also moved the server launch code into `server/mod.rs`. The only thing that took some work to upgrade was the websockets, but it's still mostly the same.

### DIFF
--- a/api/src/server/websocket.rs
+++ b/api/src/server/websocket.rs
@@ -225,20 +225,17 @@ impl StreamHandler<Result<ws::Message, ws::ProtocolError>>
 #[get("/ws/hardware/{hw_spec_slug}/programs/{program_spec_slug}/")]
 pub async fn ws_program_specs_by_slugs(
     req: HttpRequest,
-    pool: web::Data<Pool>,
-    hw_spec_slug: web::Path<String>,
-    program_spec_slug: web::Path<String>,
     stream: web::Payload,
+    pool: web::Data<Pool>,
+    params: web::Path<(String, String)>,
 ) -> Result<HttpResponse, actix_web::Error> {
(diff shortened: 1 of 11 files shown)


api/src/server/websocket.rs
```diff
@@ -174,25 +174,31 @@ impl Actor for ProgramWebsocket {
                 ctx.stop();
             } else {
                 // Not timed out, send another ping
-                ctx.ping("");
+                ctx.ping(b"");
```
JRMurr: this b string makes me mad for no reason lmao

api/src/server/websocket.rs
```diff
@@ -201,34 +207,35 @@ impl StreamHandler<ws::Message, ws::ProtocolError> for ProgramWebsocket {
...
+            Ok(ws::Message::Close(_)) => {
                 ctx.stop();
             }
-            ws::Message::Nop => {}
+
+            // Don't do anything with these messages
+            Ok(ws::Message::Binary(_))
+            | Ok(ws::Message::Continuation(_))
+            | Ok(ws::Message::Nop)
+            | Err(_) => {}
```
JRMurr: Could we unwrap msg before the match so each arm does not need the ok?

api/src/server/websocket.rs
LucasPickering: triggered by python?

api/src/server/websocket.rs
LucasPickering: Actually, I think I prefer having it this way. Otherwise, you need a nested `match`, which IMO is harder to read.

PR #34, opened 2020-02-01
Frontend setup with terminal

JRMurr:
Did create react app with typescript and got the [terminal](https://github.com/wasmerio/wasmer-js) working (thought its a little jank right now)

Making this pr before i get too far so i can make sure the structure of stuff is starting ok.

I can probably get rid of doing the class component for the terminal but i needed the onMount hook and did not want to figure out react hooks yet.

I'm gonna look into exporting the types from wasmer-js since they made the terminal in ts but did not include the type files for some reason.


frontend/.eslintrc.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
```
LucasPickering: Can you copy this file? Should have everything we want. https://github.com/LucasPickering/rps/blob/master/webapp/.eslintrc.js

frontend/public/index.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
```
LucasPickering: Can you delete the comments from this? don't want us to look like n00bs

frontend/public/manifest.json
```diff
@@ -0,0 +1,25 @@
+{
+  "short_name": "React App",
+  "name": "Create React App Sample",
```
LucasPickering: Can you update these? maybe `GDLK` and `GDLK Development Language Kit`

frontend/src/components/Terminal.tsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
+import '../../node_modules/@wasmer/wasm-terminal/lib/xterm/xterm.css';
```
LucasPickering: You shouldn't have to do this -- webpack should be set up so that you can just do `import @wasmer/wasm-terminal/lib/xterm/xterm.css`

frontend/src/components/Terminal.tsx
```diff
@@ -0,0 +1,58 @@
...
+  } else if (commandName === 'echo') {
+    const callbackCommand = async (options: CommandArgs): Promise<string> => {
+      // first index is the 'echo' ignore that
+      return options.args.slice(1).join(' ');
+    };
+    return callbackCommand;
+  }
+};
+
+export default class Terminal extends React.Component<{}> {
```
LucasPickering: Can you make this a functional component?

frontend/src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
LucasPickering: do we need this file?

frontend/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
```
LucasPickering: Can you include `"baseUrl": "src/"` in here? That should let us do absolute imports of our own files.

PR #35, opened 2020-02-02
Add set_content and delete file operations

LucasPickering:
Added two write operations: `set_content` and `delete`. Will do create and rename later. Also fixed a bug cause I forgot to implement `exists` for the program source node handler, which meant it always returned `true` for `exists()`.

### DIFF
--- a/api/src/util.rs
+++ b/api/src/util.rs
@@ -13,7 +13,7 @@ pub type PooledConnection =
 #[cfg(test)]
 mod tests {
     use super::*;
-    use diesel::{Connection, PgConnection};
+    use diesel::Connection;
 
     /// Helper to create a database connection for testing. This establishes
     /// the connection, then starts a test transaction on it so that no changes
@@ -30,5 +30,17 @@ mod tests {
 
         (&conn as &PgConnection).begin_test_transaction().unwrap();
         conn
+    }
+
+    /// Assert that the first value is an Err, and that its string form matches
+    /// the second argument.
+    #[macro_export]
+    macro_rules! assert_err {
+        ($res:expr, $msg:tt $(,)?) => {
+            match $res {
+                Ok(_) => panic!("Expected Err, got Ok"),
+                Err(err) => assert_eq!(format!("{}", err), $msg),
+            }
+        };
     }
 }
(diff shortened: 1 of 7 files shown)


api/src/util.rs
```diff
@@ -31,4 +31,16 @@ mod tests {
...
     }
+
+    /// Assert that the first value is an Err, and that its string form matches
+    /// the second argument.
+    #[macro_export]
+    macro_rules! assert_err {
+        ($res:expr, $msg:tt $(,)?) => {
+            match $res {
+                Ok(_) => panic!("Expected Err, got Ok"),
+                Err(err) => assert_eq!(format!("{}", err), $msg),
```
JRMurr: is there a generic `fail!` or something like that so we dont have to panic?

api/src/vfs/program.rs
```diff
@@ -153,4 +171,59 @@ impl VirtualNodeHandler for ProgramSourceNodeHandler {
...
+        // Diesel doesn't support updates with joins, so we have to fetch the
+        // ID of the UserProgram first, then do the update in a second query.
+        let user_program_id: i32 = UserProgram::filter_by_file_name(
+            context.user.id,
+            hw_spec_slug,
+            program_spec_slug,
+            path_segment,
+        )
+        .select(id)
+        .get_result(context.conn())?;
```
JRMurr: Can you do subqueries or with clauses so its still executed as one query?

JRMurr:
So like what would a mutation query the client sends look like?

api/src/vfs/program.rs
LucasPickering: idk but I don't think it's worthing trying to figure out. I don't think the performance impact is going to be much at all, the SELECT will be fast.

api/src/util.rs
LucasPickering: That's exactly what `panic!` is. It indicates an internal bug in a pretty unlikely scenario.

LucasPickering:
> 
> 
> So like what would a mutation query the client sends look like?

Mutation:

```
mutation {
  fsNode(username: "user1", path: "/hardware/hw1/programs/prog1/program.gdlk") {
    setContent(content: "READ\nWRITE\nREAD\nWRITE\n") {
      name
      content
    }
  }
}
```

Response:
```
{
  "data": {
    "fsNode": {
      "setContent": {
        "name": "program.gdlk",
        "content": "READ\nWRITE\nREAD\nWRITE\n"
      }
    }
  }
}
```

PR #36, opened 2020-02-04
Clean up file system permissions

LucasPickering:
A few changes to fs permissions:
- Removed execute permission
  - I don't see any need for it, since all the executable commands won't actually be defined in the FS, they'll just be on the frontend. And for directories, I think it just complicates it and makes it harder to build, and harder to learn if you don't already understand unix permissions.
- Added read permission checks to `content` and `children`
- Unrelated, but removed the `slice_patterns` from the list of our nightly features, since it got stabilized in 1.42

### DIFF
--- a/api/src/vfs/mod.rs
+++ b/api/src/vfs/mod.rs
@@ -150,7 +150,7 @@ impl Node {
     }
 
     /// Gets the actual data of this node. This is only possible for files, and
-    /// requires read permission. Return an error if this node is a directory,
+    /// requires read permission. Returns an error if this node is a directory,
     /// or does not have read permission.
     pub fn content(&self) -> Result<String> {
         match self.vnode.node_type {
@@ -455,7 +455,7 @@ impl VirtualNodeHandler for SimpleDirHandler {
 
 /// The entire VFS node tree. This defines the layout of the tree. Each node has
 /// a path spec, a handler that defines how it behaves, and optionally children.
-/// Obviously, only directions can have children.
+/// Obviously, only directories can have children.
 static VFS_TREE: VirtualNode = VirtualNode::dir(
     // If you update something here, make sure to update the comment above!
     SegmentSpec::Fixed(""),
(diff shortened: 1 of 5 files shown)


api/src/vfs/mod.rs
```diff
@@ -104,25 +149,31 @@ impl Node {
         )
     }
 
-    /// Gets the actual data of this node. This is only possible for files. If
-    /// called for a directory, will return
-    /// <ServerError::UnsupportedFileOperation>.
+    /// Gets the actual data of this node. This is only possible for files, and
+    /// requires read permission. Will return an error if this node is a
```
JRMurr: I know im not a grammar expert buttt starting a sentence with will feels weird lol

api/src/vfs/mod.rs
```diff
@@ -104,25 +149,31 @@ impl Node {
...
+            }
             // Can't do this for directories, ya dummy
             NodeType::Directory => Err(ServerError::UnsupportedFileOperation),
         }
     }
 
     /// Gets a list of each child of this node. This is only possible for
-    /// directories, as files have no children. If called for a file, will
-    /// return <ServerError::UnsupportedFileOperation>.
+    /// directories, as files have no children. Requires read permission. Will
```
JRMurr: same will sentence start

api/src/vfs/mod.rs
```diff
@@ -194,15 +244,12 @@ impl Node {
...
-        if self.permissions()?.write {
-            self.vnode.handler.delete(
-                &self.context,
-                &self.path_variables,
-                &self.path_segment,
-            )
-        } else {
-            Err(ServerError::PermissionDenied)
-        }
+        self.permissions()?.require_write()?; // Require write perms
```
JRMurr: probs can get rid of the `// Require write perms` comment

api/src/vfs/mod.rs
```diff
@@ -150,12 +150,12 @@ impl Node {
     }
 
     /// Gets the actual data of this node. This is only possible for files, and
-    /// requires read permission. Will return an error if this node is a
-    /// directory, or does not have read permission.
+    /// requires read permission. Return an error if this node is a directory,
+    /// or does not have read permission.
```
JRMurr: returns

PR #37, opened 2020-02-04
Improve CI speed

LucasPickering:
Shaved a couple minutes off the CI run time with a few changes:

- Removed cargo-watch from the CI build
- Removed `libssl-dev` and `wget` (Don't need either)

Also a few other changes:

- Upgraded nightly again because clippy was having memory issues on the old version
- Disabled a lint that apparently was enabled in new nightly, because Diesel's macros violate it

I tried the caching but not too much. Honestly it doesn't seem like it's actually working for the other steps and it's harder to do with docker. I'm planning on reworking this all at some point when we start deploying this thing, so I'll try it again then.


LucasPickering:
It's saying the checks failed because I cancelled an older run. You can see it passed on the latest commit, so idk what's up.

PR #38, opened 2020-02-05
Add createChild operation to file system

LucasPickering:
- Added `create_child` operation
- Exposed that operatio as `createChild` in GQL
- Renamed `Node` to `NodeReference`
  - I realized that a `Node` is really a pointer to to an actual node, and now with the new code we have `Node`s that don't yet have corresponding physical nodes, so I think `NodeReference` is a better name
- Removed delete functionality for directories. Right now we don't have any dirs that the user will be able to delete anyway, and the permissioning rules I picked wouldn't work. I think what we'll end up wanting to do is make it so that delete permissions work like create, where you need write permissions on the _parent_ and the node itself to be able to delete.

Example query/response:

```graphql
mutation {
  fsNode(username: "user1", path: "/hardware/hw1/programs/prog1") {
    createChild(name: "program.gdlk") {
      name
      content
    }
  }
}
```

```json
{
  "data": {
    "fsNode": {
      "createChild": {
        "name": "program.gdlk",
        "content": ""
      }
    }
  }
}
```

### DIFF
--- a/api/src/error.rs
+++ b/api/src/error.rs
@@ -30,6 +30,11 @@ pub enum ServerError {
     /// perform the attempted operation.
     #[fail(display = "Permission denied")]
     PermissionDenied,
+
+    /// User attempted to create a node, but a node already exists at that
+    /// path.
+    #[fail(display = "Node already exists")]
+    AlreadyExists,
 }
 
 impl From<r2d2::Error> for ServerError {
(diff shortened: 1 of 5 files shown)


JRMurr:
@LucasPickering we doing squash now or still rebasing?

PR #41, opened 2020-02-15
Improve parsing

LucasPickering:
Rewrote some parsing logic to try to make it easier to follow.


core/src/parse.rs
```diff
@@ -232,98 +226,115 @@ fn jump_stmt(input: &str) -> ParseResult<'_, Statement> {
...
-fn comment_or_spaces(input: &str) -> ParseResult<'_, &str> {
-    let (input, _) =
-        many0(tuple((space0, many1(alt((comment, line_ending))), space0)))(
-            input,
-        )?;
-    Ok((input, ""))
+/// Parses one source statement. This consumes surrounding spaces/comments, but
+/// not the line ending.
+fn statement(input: &str) -> ParseResult<'_, Statement> {
+    context("Statement", alt((label_stmt, operator_stmt, jump_stmt)))(input)
```
JRMurr: I think the comment doesn't match. Doesn't look like the parsers in the alt consume surrounding spaces/comments

PR #42, opened 2020-02-15
update versions for frontend

JRMurr:
why webpack make me sad


.nvmrc
```diff
@@ -1 +1 @@
-12.14.1
\ No newline at end of file
+12.16.0
```
LucasPickering: move to frontend pls

PR #45, opened 2020-02-17
Rewrite AST and parsing to include source span data

LucasPickering:
big chungus


core/src/consts.rs
```diff
@@ -1,10 +1,3 @@
-/// The name of the register that holds the current input length
-pub const REG_INPUT_LEN: &str = "RLI";
-/// The prefix for all registers that hold stack lengths (e.g. "RS0", "RS1")
-pub const REG_STACK_LEN_PREFIX: &str = "RS";
-/// The prefix for all user registers (e.g. "RX0")
-pub const REG_USER_PREFIX: &str = "RX";
```
LucasPickering: There aren't needed anymore since we're not using them to render errors anymore

core/src/parse.rs
```diff
@@ -96,177 +306,27 @@ where
...
-                .into_iter()
-                .chain(iter::once(last_line.unwrap_or(None)))
-                .filter_map(std::convert::identity)
-                .collect()
-        },
+fn line(input: RawSpan) -> ParseResult<'_, Option<SpanNode<Statement<Span>>>> {
+    delimited(
+        space0,
+        opt(Statement::parse_node),
+        tuple((space0, opt(line_comment))),
```
JRMurr: Im a little confused on how this handles the top level `alt` we had before. Does Nom locate just kinda handle that based on how the traits are implemented

JRMurr: nvm just noticed the alt was in the trait impl for statement

core/src/parse.rs
```diff
@@ -280,350 +340,570 @@ fn parse(input: &str) -> Result<Vec<Statement>, CompileError> {
...
+            )
+            .unwrap()
+            .body,
+            vec![
+                Node(
+                    Statement::Label(Node(
+                        LabelDecl("LBL".into()),
+                        span(2, 17, 2, 21)
+                    )),
+                    span(2, 17, 2, 21)
```
JRMurr: Is this the main example of duplication you were talking about?

PR #48, opened 2020-02-25
Optimize docker builds and CI speed

LucasPickering:
Reworked the dockerfiles so that they don't contain our source, just the environment setup that's needed. That means they also don't contain any prebuilt/preinstalled deps. With proper caching in the CI though, that shouldn't be an issue for speed. The big different here is that we're now pushing the images and pulling them down during CI, to prevent having to re-build them.


api/Dockerfile
```diff
@@ -1,26 +1,22 @@
...
     libpq-dev \
     pkg-config \
     && \
     rm -rf /var/lib/apt/lists/*
 
-ADD https://github.com/jwilder/dockerize/releases/download/$DOCKERIZE_VERSION/dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz .
-RUN tar -C /usr/local/bin -xzvf dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz && \
-    rm dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz
+RUN curl -sSL https://github.com/jwilder/dockerize/releases/download/${DOCKERIZE_VERSION}/dockerize-alpine-linux-amd64-$DOCKERIZE_VERSION.tar.gz \
+    | tar xzv -C /usr/local/bin
```
LucasPickering: Turns out it's better to do this because you end up with smaller images.

PR #49, opened 2020-02-28
Random tiny hygiene stuff

LucasPickering:
- Started using the re-exported versions of `validator` and `validator_derive` in the API crate so that we don't have to re-specify the dependency. I don't _think_ this will make any difference on build time because it probably wasn't building it twice, but there's a chance this actually speeds up builds.
- Separate the development CLI into another crate, just to clean out the core crate as much as possibly. Should make building core faster now since we don't have to build structopt and serde_json.
- Clean up .gitignore a tiny bit
- Use read-only permissionf ro mounted local dirs in `docker-compose.yml`, so that the containers don't accidentally overwrite files.


JRMurr:
@LucasPickering conflicts

PR #52, opened 2020-03-02
Better parse errors

LucasPickering:
Rewrote a lot of parse code to get this to work. Basically what it's doing is, when there's an error, it goes down the error chain and finds the first error with context, and uses the context to form the error message. We're not using `convert_error` anymore, cause it's kinda shit. Example:

### Source
```
READ XR0
```

### Output
```
Syntax error at 1:6: Expected register reference
    | 
  1 | READ XR0
    |      ^
    | 
```

Take a look at the tests for more examples. Lemme know if you think of any possible parse errors that I didn't cover.

### DIFF
--- a/core/src/parse.rs
+++ b/core/src/parse.rs
@@ -5,24 +5,20 @@ use crate::{
         StackRef, UserRegisterId, ValueSource,
     },
     error::{CompileError, SourceErrorWrapper, WithSource},
-    util::Span,
+    util::{RawSpan, Span},
     Compiler,
 };
 use nom::{
     branch::alt,
     bytes::complete::{is_not, tag, tag_no_case, take_while1},
     character::complete::{char, digit1, line_ending, space0, space1},
-    combinator::{all_consuming, cut, map, map_res, opt, recognize},
-    error::{context, convert_error, ParseError, VerboseError},
-    lib::std::ops::RangeTo,
+    combinator::{all_consuming, cut, map, map_res, opt, peek, recognize},
+    error::{context, ErrorKind, ParseError, VerboseError, VerboseErrorKind},
     multi::many0,
-    sequence::{delimited, preceded, separated_pair, terminated, tuple},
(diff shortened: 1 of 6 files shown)


core/tests/compile_error.rs
```diff
@@ -1,41 +1,101 @@
...
+                .unwrap_err()
+                .errors()
+                .iter()
+                .map(|err| format!("{}", err))
+                .collect();
+        let strs: Vec<&str> =
+            actual_errors.iter().map(String::as_str).collect();
+        assert_eq!(strs.as_slice(), $expected_errors);
+    };
+}
```
LucasPickering: I switched to macros cause then the line that triggers the error is in the test, instead of always just pointing to the same line in the helper function.

x.py
```diff
@@ -162,10 +162,7 @@ def test_api(self, debug, tests, backtrace):
         db_url = f"postgres://root:root@db/{API_TEST_DB}"
         run_in_docker_service(
             API_SERVICE,
-            ["diesel", "migration", "run"],
-            # TODO switch to this version once we have enough tables that
-            # diesel and rustfmt stop fighting over schema.rs
-            # ["diesel", "migration", "run", "--locked-schema"],
+            ["diesel", "migration", "run", "--locked-schema"],
```
LucasPickering: Unrelated, but I noticed we can do this now

core/src/parse.rs
```diff
@@ -92,32 +75,26 @@ impl<'a> Parse<'a> for LabelDecl {
...
-                map(
-                    preceded(tag_no_case("RX"), cut(UserRegisterId::parse)),
-                    RegisterRef::User,
-                ),
-            )),
-        )(input)
+        alt((
+            // "RLI" => RegisterRef::InputLength
+            map(tag_no_case("RLI"), |_| RegisterRef::InputLength),
+            // "RSx" => RegisterRef::StackLength(x)
```
JRMurr: Don't need to address this in this PR but if they did `RL` or some bad tag im not sure what kind of error they would get

PR #60, opened 2020-04-09
Replace rust-based seeding with sql script

LucasPickering:
The big buff here is that we can remove `structopt` from the API crate, which brings the build time down by ~1 min (on my machine), which is like 40%. Also, it now runs the seed script during CI, so it can't get out of sync with the schema.


api/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN apt-get update && \
     curl \
     libpq-dev \
     pkg-config \
+    postgresql \
```
LucasPickering, line 11: Had to add this so we can use the `psql` shell in the CI. It's super small, less than 1MB.

PR #61, opened 2020-04-12
Infrastructure changes to get ready for API tests

LucasPickering:
- Split the `gdlk_api` crate into both a bin and a lib. We can't write integration tests for bin crates, so we need the lib crate in order to have the API tests.
- Changed the DB setup a bit. Instead of having a SQL script that we run for each DB to enable the UUID extension, instead one script runs when the container starts to enable that extension for the `template1` DB. Then, any DB we create after that automatically has the extension enabled. This makes it easier to set up the test DB.
- Wrote a couple tests to get the approach worked out. It runs the tests in serial and wipes all tables after each run to make sure the tests don't fight each other.


api/Cargo.toml
```diff
@@ -18,11 +18,20 @@ failure = "0.1"
 gdlk = { path = "../core" }
 juniper = "0.14.2"
 juniper-from-schema = "0.5.2"
-lazy_static = "1.4"
 r2d2 = "0.8"
-regex = "1.3"
```
LucasPickering, line 23: We weren't using either of these anymore

api/Cargo.toml
```diff
@@ -18,11 +18,20 @@ failure = "0.1"
...
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 uuid = "^0.7.0"
 validator = "0.10.0"
 validator_derive = "0.10.0"
+
+[dev-dependencies]
+maplit = "^1.0.0"
```
LucasPickering, line 29: Needed just for tests

api/tests/test_graphql.rs
```diff
@@ -0,0 +1,198 @@
...
+    fn drop(&mut self) {
+        let conn = self.context.get_db_conn().unwrap();
+        delete_tables!(
+            &conn,
+            models::User,
+            models::HardwareSpec,
+            models::ProgramSpec,
+            models::UserProgram,
+            // Any new table needs to be added here!
+        );
```
LucasPickering, line 82: We need a way to have the tests not affect each other via the DB. I tried transactions but that doesn't work because the API uses a connection pool and each transaction is attached to one connection. This isn't a great solution since we have to maintain this list, but it's the best I can come up with.

JRMurr: Could we limit the connection pool for the tests so transactions work?

JRMurr: Still this drop trait is so lit that it just happen automagically

PR #62, opened 2020-04-14
Use base64 encoding for cursors

LucasPickering:
- base64 cursors instead of just stringified ints
- Added two more hardware specs to the seed data
- Cleaned up API errors a bit
  - I'm gonna be doing more work on this soon. We have a few other places that currently return 400-ish errors when they should be 404-ish
- Fixed a couple other API bugs
  - `totalCount` field was broken (I'm surprised diesel didn't catch this error statically)
  - Changed `username` param to `userId` on a couple fields, since they actually expect an ID

### Query

```graphql
{
  hardwareSpecs {
    totalCount
    pageInfo {
      startCursor
      endCursor
      hasNextPage
      hasPreviousPage
    }
    edges {
      cursor
      node {
        id
        slug
        programSpecs {
          edges {
            cursor
            node {
              id
              slug
            }
          }
        }
      }
    }
  }
}
```

### Response

```json
{
  "data": {
    "hardwareSpecs": {
      "totalCount": 3,
      "pageInfo": {
        "startCursor": "AAAAAA==",
        "endCursor": "AAAAAg==",
        "hasNextPage": false,
        "hasPreviousPage": false
      },
      "edges": [
        {
          "cursor": "AAAAAA==",
          "node": {
            "id": "57939ed7-e0bd-436c-a540-7e4770facd2a",
            "slug": "hw1",
            "programSpecs": {
              "edges": [
                {
                  "cursor": "AAAAAA==",
                  "node": {
                    "id": "80477d00-89d8-436f-8cdf-3fb4257865f3",
                    "slug": "prog1"
                  }
                },
                {
                  "cursor": "AAAAAQ==",
                  "node": {
                    "id": "6610fbf6-1f90-422b-a851-4ebc7d3a03f9",
                    "slug": "prog2"
                  }
                }
              ]
            }
          }
        },
        {
          "cursor": "AAAAAQ==",
          "node": {
            "id": "4ba1fc7f-926c-43a0-a4f5-bde7f64d50f1",
            "slug": "hw2",
            "programSpecs": {
              "edges": []
            }
          }
        },
        {
          "cursor": "AAAAAg==",
          "node": {
            "id": "a7595041-f24b-429c-876f-6b421e916a3e",
            "slug": "hw3",
            "programSpecs": {
              "edges": []
            }
          }
        }
      ]
    }
  }
}
```

### DIFF
--- a/api/src/server/gql/hardware_spec.rs
+++ b/api/src/server/gql/hardware_spec.rs
@@ -139,7 +139,7 @@ impl HardwareSpecConnection {
 
     fn get_total_count(&self, context: &Context) -> ServerResult<i32> {
         match hardware_specs::table
-            .select(dsl::count(dsl::count_star()))
+            .select(dsl::count_star())
             .get_result::<i64>(&context.get_db_conn()?)
         {
             // Convert i64 to i32 - if this fails, we're in a rough spot
--- a/api/src/server/gql/hardware_spec.rs
+++ b/api/src/server/gql/hardware_spec.rs
@@ -143,7 +143,7 @@ impl HardwareSpecConnection {
             .get_result::<i64>(&context.get_db_conn()?)
         {
             // Convert i64 to i32 - if this fails, we're in a rough spot
-            Ok(count) => Ok(count.try_into()?),
+            Ok(count) => Ok(count.try_into().unwrap()),
             Err(err) => Err(err.into()),
         }
     }
--- a/api/src/server/gql/mod.rs
+++ b/api/src/server/gql/mod.rs
@@ -16,9 +16,11 @@ use diesel::{
     pg::upsert, ExpressionMethods, OptionalExtension, QueryDsl, RunQueryDsl,
     Table,
 };
+use failure::Fallible;
 use gdlk::Valid;
 use juniper_from_schema::graphql_schema_from_file;
-use std::sync::Arc;
+use serde::{Serialize, Serializer};
+use std::{convert::TryInto, sync::Arc};
 use uuid::Uuid;
 use validator::{Validate, ValidationError, ValidationErrors};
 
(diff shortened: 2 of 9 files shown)


api/src/server/gql/mod.rs
```diff
@@ -104,35 +106,43 @@ impl QueryFields for Query {
...
+    let mut error: ValidationError = match cursor.to_index() {
+        Ok(index) if index >= 0 => {
+            return Ok(());
+        }
+        _ => ValidationError::new("cursor"),
+    };
+    error.add_param("message".into(), &"Invalid cursor value");
+    Err(error)
+}
+
```
LucasPickering, line 145: This fixes the nasty cursor validation error format from before. I had to manually implement `Serialize` for `Cursor` since the struct is generated by the GraphQL macro above, so I can't put `#[derive(Serialize)]` on it. Here's the improvement:

### Query

```graphql
{
  hardwareSpecs(after: "bad-cursor") {
    totalCount
  }
}
```

### Response Before

```json
{
  "data": null,
  "errors": [
    {
      "message": "Input validation error(s)",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "hardwareSpecs"
      ],
      "extensions": {
        "after": {
          "__all__": [
            {
              "cursor": "\"bad-cursor\""
            }
          ]
        }
      }
    }
  ]
}
```

### Response After

```json
{
  "data": null,
  "errors": [
    {
      "message": "Input validation error(s)",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "hardwareSpecs"
      ],
      "extensions": {
        "after": [
          {
            "message": "\"Invalid cursor value\"",
            "value": "\"bad-cursor\""
          }
        ]
      }
    }
  ]
}
```

api/src/server/gql/hardware_spec.rs
```diff
@@ -139,11 +139,11 @@ impl HardwareSpecConnection {
 
     fn get_total_count(&self, context: &Context) -> ServerResult<i32> {
         match hardware_specs::table
-            .select(dsl::count(dsl::count_star()))
+            .select(dsl::count_star())
             .get_result::<i64>(&context.get_db_conn()?)
         {
             // Convert i64 to i32 - if this fails, we're in a rough spot
-            Ok(count) => Ok(count.try_into()?),
+            Ok(count) => Ok(count.try_into().unwrap()),
```
LucasPickering, line 146: I changed these errors to `unwrap`s cause if they happen, it means the table is too big for its length to fit in an `i32`. I think we want a panic to occur in these cases, so that they get logged and we can potentially get notified about them. I'm planning on adding logic soon to catch panics in response handlers, so this won't crash the server.

PR #67, opened 2020-04-30
Compile in frontend

JRMurr:


frontend/tsconfig.json
```diff
@@ -16,5 +16,8 @@
     "jsx": "react",
     "baseUrl": "src/"
   },
-  "include": ["src"]
+  "include": [
+    "src/**/*",
+    "node_modules/@wasmer/wasm-terminal/lib/optimized/packages/wasm-terminal/src/index.d.ts"
```
LucasPickering, line 21: Can we revert these changes since we're not using the terminal?

frontend/package.json
```diff
@@ -30,8 +30,7 @@
     "@types/relay-runtime": "^8.0.8",
     "@typescript-eslint/eslint-plugin": "^2.29.0",
     "@typescript-eslint/parser": "^2.29.0",
-    "@wasm-tool/wasm-pack-plugin": "^1.3.1",
-    "@wasmer/wasm-terminal": "^0.9.1",
+    "@wasm-tool/wasm-pack-plugin": "^1.1.0",
```
LucasPickering, line 33: Did you downgrade this package intentionally?

wasm/Cargo.lock
```diff
@@ -1,30 +1,159 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
+[[package]]
```
LucasPickering, line 3: We should be able to delete this whole file now that this package is part of the workspaces.

wasm/src/lib.rs
```diff
@@ -1,12 +1,29 @@
...
+    let hw_spec: HardwareSpec = Default::default();
+    let valid_spec = match Valid::validate(hw_spec) {
+        Ok(spec) => spec,
+        _ => return "ERROR: Invalid Hardware Spec".to_string(),
+    };
+    match Compiler::compile(source.to_string(), valid_spec) {
+        Ok(prog) => format!("{:?}", prog),
+        Err(err) => format!("{:#}", err),
+    }
 }
```
LucasPickering, line 16: Can you make this take in a `HardwareSpec` and return the AST object?

.github/workflows/ci.yml
```diff
@@ -115,6 +115,33 @@ jobs:
...
+  #     - name: Cache cargo build
+  #       uses: actions/cache@v1
+  #       with:
+  #         path: target
+  #         key: ${{ runner.os }}-cargo-build-target-${{ hashFiles('Cargo.lock') }}
+  #     - name: Run test
+  #       uses: actions-rs/cargo@v1
+  #       with:
+  #         command: test
+  #         args: -p gdlk_wasm
```
LucasPickering, line 142: Can you just remove this? Since it's just a copy of the core test job, it won't be hard to re-create if we write tests

wasm/src/lib.rs
```diff
@@ -1,12 +1,17 @@
...
+        _ => return JsValue::from_str("ERROR: Invalid Hardware Spec"),
+    };
 
-#[wasm_bindgen]
-pub fn greet(name: &str) -> String {
-    format!("suh, {}!", name)
+    return match Compiler::compile(source.to_string(), valid_spec) {
+        Ok(res) => JsValue::from_serde(&res.program()).unwrap(),
+        Err(err) => JsValue::from_serde(&err).unwrap(),
+    };
```
LucasPickering, line 16: can you get rid of the `return`?

core/src/lib.rs
```diff
@@ -107,6 +107,10 @@ impl Compiler<Program<Span>> {
     pub fn allocate(self, program_spec: &Valid<ProgramSpec>) -> Machine {
         Machine::new(&self.hardware_spec, program_spec, self.ast, self.source)
     }
+
+    pub fn program(self) -> Program<Span> {
```
LucasPickering, line 111: Can you add a short comment here?

JRMurr:
I had to be a little jank with the compiler ts wrapper since we need the async import. Its a static class that will need to have init called on it at some point. You would use it like this 
```javascript
Compiler.init().then(() => {
    const hwSpec = Compiler.makeHardwareSpec(10, 10, 10);
    console.log(
      JSON.stringify(Compiler.compile('READ RX0\n', hwSpec), null, 4)
    );
});
```

PR #69, opened 2020-05-01
Set up cargo-make

LucasPickering:
Sets up [cargo-make](https://github.com/sagiegurari/cargo-make), which is a task runner. The idea is to replace `x.py` with this. I think this is better cause it's less code to maintain, and since it runs on Rust, we can use it in inside our docker images more easily. I left `x.py` in place for now, once we're happy that we're not missing any functionality, we can delete it.


JRMurr:
@LucasPickering conflicts

PR #72, opened 2020-05-08
Add IDE and minimal workflow to get there

LucasPickering:
Big ol' PR to add some basic UI views. Three new views:

### Home page (list of hardware specs)

image

### Program Spec page (list of program solutions)

image

### IDE

image

The IDE is fully functional, and you can save programs, compile them, and step through them. There's no client-side compilation yet, it's all on the FE.

You might wanna check out this branch and test it out yourself. A lot of styling still needs to be done, this is mostly just functionality.


JRMurr:
Just some initial feedback as i mess around

In the editor it would be nice to see the input and expected output before you hit compile.
Also some way to go back to the list of solutions for the program would be nice. Right now you have to hit back in the browser

LucasPickering:
Yeah I definitely agree on both of those. For the I/O, I've implemented that in my next branch already. I want to pre-render the registers and stacks as well, since that information is available via the hardware spec, but I need the wasm stuff in order to do it.
For the back button, I was gonna include it in the header somehow but I haven't thought about it too much yet. I'll get to it at some point tho.

PR #73, opened 2020-05-14
More IDE improvements

LucasPickering:
- Switched to using a lib called Ace for the editor, which is what a lot of online code editors use. it's nut
- Moved the interval stepping logic to the backend. Previously it ran an interval on the frontend that would send a "step" msg every x ms, then the server would respond with the new machine state. If x got too small though, it couldn't keep up. Now the client just sends one message saying "step a x ms" and the server maintains the interval loop itself, and emits a new machine state each time.
  - This made the speed better, but it's still slower than I'd like. We might have to switch to your original idea of executing it on the frontend entirely, then execute it again on the backend to verify the results for leaderboard purposes. I think once the wasm stuff is working, this should be pretty easy to do, and should reduce the load on the server which is neat (maybe even eliminate the need for websockets? that would be dope)
- Made some various theme/design tweaks around the site. Let me know how you like the color scheme.

Here's what the IDE looks like now:
image

Again, you'll prob want to run it yourself to test all the shit out.


frontend/src/hooks/useWebSocket.ts
```diff
@@ -32,7 +33,7 @@ interface SocketCallbacks<In, Out> {
  * `Out` - The type of data sent out over the websocket.
  */
 const useWebSocket = <In, Out>(
-  addr: string,
+  address: string | undefined,
```
LucasPickering, line 36: I just tweaked the logic so that if you pass `undefined` as the address, it doesn't connect.

LucasPickering:
Here's a diagram of the change I made to the stepper.

## Old

|Client| |Server
|-|-|-|
||Step -->||
||<-- State||
||Step -->||
||<-- State||

So each step required sending a message in both directions. When I tried to run it every 50ms, it couldn't keep up all the time, and it would only be worse over the internet.

## New

|Client||Server|
||AutoStep(50 ms) -->||
||<-- State||
||<-- State||
||<-- State||

So now that the 50ms loop is managed on the server, you don't need a full round trip on each step. This got the speed a lot faster. It still stutters sometimes though, so that's why I'm thinking we might want to just execute it on the frontend at some point.

PR #75, opened 2020-05-20
Implement wasm compilation/execution in IDE

LucasPickering:
Lots of shit here. Compilation and execution in the frontend is working. There are some bugs, and this also leaves behind a lot of dead code. I'm going to clean up all that stuff next.

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -25,8 +25,10 @@ r2d2 = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 uuid = "^0.7.0"
-validator = "0.10.0"
-validator_derive = "0.10.0"
+# validator = "0.10.0"
+# validator_derive = "0.10.0"
+validator = { git = "https://github.com/LucasPickering/validator", branch = "blanket-ref-impl" }
+validator_derive = { git = "https://github.com/LucasPickering/validator", branch = "blanket-ref-impl" }
 
 [dev-dependencies]
 maplit = "^1.0.0"
--- a/api/src/server/mod.rs
+++ b/api/src/server/mod.rs
@@ -1,7 +1,7 @@
 //! All code related to the webserver. Basically anything that calls Actix
 //! lives here.
 mod gql;
-mod websocket;
+// mod websocket;
 
 pub use crate::server::gql::{create_gql_schema, Context, GqlSchema};
 use crate::util::Pool;
@@ -54,7 +54,7 @@ pub async fn run_server(pool: Pool, host: String) -> io::Result<()> {
             // routes
             .service(route_graphql)
             .service(route_graphiql)
-            .service(websocket::ws_program_specs_by_slugs)
+        // .service(websocket::ws_program_specs_by_slugs)
     })
     .bind(host)?
     .run()
--- a/core/src/machine.rs
+++ b/core/src/machine.rs
@@ -1,3 +1,5 @@
+#[cfg(feature = "wasm")]
+use crate::ast::wasm::{LangValueArrayMap, LangValueMap, SourceElement};
 use crate::{
     ast::{
         compiled::{Instruction, Program},
@@ -13,6 +15,8 @@ use crate::{
 use std::{
     cmp::Ordering, collections::HashMap, convert::TryInto, iter, num::Wrapping,
 };
+#[cfg(feature = "wasm")]
+use wasm_bindgen::{prelude::*, JsCast};
 
 /// A steppable program executor. Maintains the current state of the program,
 /// and execution can be progressed one instruction at a time.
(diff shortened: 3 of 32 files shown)


api/Cargo.toml
```diff
@@ -25,8 +25,10 @@ r2d2 = "0.8"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 uuid = "^0.7.0"
-validator = "0.10.0"
-validator_derive = "0.10.0"
+# validator = "0.10.0"
+# validator_derive = "0.10.0"
+validator = { git = "https://github.com/LucasPickering/validator", branch = "blanket-ref-impl" }
+validator_derive = { git = "https://github.com/LucasPickering/validator", branch = "blanket-ref-impl" }
```
LucasPickering, line 31: I have a PR up to this repo, once that gets merged and released we can go back to the stable version

api/src/server/mod.rs
```diff
@@ -1,7 +1,7 @@
 //! All code related to the webserver. Basically anything that calls Actix
 //! lives here.
 mod gql;
-mod websocket;
+// mod websocket;
```
JRMurr, line 4: Don't want to remove this just yet?

core/src/machine.rs
```diff
@@ -356,9 +372,82 @@ impl Machine {
...
+    pub fn wasm_registers(&self) -> LangValueMap {
+        // Convert the keys of the register map to strings
+        let regs_by_name: HashMap<String, LangValue> = self
+            .registers()
+            .into_iter()
+            .map(|(reg_ref, reg_value)| (reg_ref.to_string(), reg_value))
+            .collect();
+        // Convert the hashmap to a js object. Be careful here!
+        JsValue::from_serde(&regs_by_name).unwrap().unchecked_into()
+    }
```
JRMurr, line 413: We talked about this a bit but for js-val stuff did you just decide to handle the types in ts instead of adding that custom ts type export?

frontend/src/components/ide/ProgramIde.tsx
```diff
@@ -69,115 +71,189 @@ const useLocalStyles = makeStyles(({ palette, spacing }) => {
...
-    (newSource: string): void => {
-      // Any time we change the source code, we want to invalidate the compiled
-      // program.
-      setMachineState(undefined);
-      setSourceCode(newSource);
+
+  // This wasm value is NOT safe to share outside this component. It's stored
+  // in a ref because it contains pointers, which often don't reflect changed
+  // values to React. As such, making it a state field would be useless. This
+  // valu ehas to be manually transformed into compiledState after changes.
```
JRMurr, line 114: valu ehas

frontend/src/util/compile.ts
```diff
@@ -1,37 +1,151 @@
...
+ * @param value The value to check
+ * @returns true if the value is a SourceElement, false if not
+ */
+function isSourceElement(value: object): value is SourceElement {
+  // TODO make this more robust after
+  // https://github.com/Microsoft/TypeScript/issues/21732
+  return 'text' in value && 'span' in value;
+}
+
+export interface MachineState {
```
JRMurr, line 28: Could we export the struct for `Machine` and use that as a type here?

frontend/src/util/compile.ts
```diff
@@ -1,37 +1,151 @@
...
-  ): HardwareSpec => {
-    return new Compiler.gdlk.HardwareSpec(
-      num_registers,
-      num_stacks,
-      max_stack_length
-    );
-  };
+      return {
+        type: 'compiled',
+        // Do NOT try to change this to a spread!
```
JRMurr, line 138: ?

api/src/server/mod.rs
LucasPickering: nah, i will in my next pr

core/src/machine.rs
LucasPickering: I did add the custom types, look at the return type of the func. These types are defined in `ast::wasm`, and they just map to string version of ts types. So there isn't any actual type checking, but it has the type assigned in TS.

frontend/src/util/compile.ts
LucasPickering: unfortunately no, it doesn't really work if we use the `Machine` type in the react code. the issue is that react re-renders when it sees that props/context have changed. but with the `Machine` wasm type, as far as react is concerned, the value is just a pointer. So the program counter/input/output etc in the machine might change, but the pointer doesn't change so react never rerenders. we need this wrapping logic to get react to re-render after changes.

frontend/src/util/compile.ts
LucasPickering: ah this makes less sense now, i'll remove the comment

PR #77, opened 2020-05-21
New lang features: DIV and RZR

LucasPickering:
- Add the `DIV` instruction, which does flooring division
  - Also includes a new `DivideByZero` runtime error
- Added the `RZR` instruction (for register-zero), which emulates `/dev/null`
  - Reading from it always gives you zero
  - You can write to it, but your value just disappears
- Fixed a register/stack reference parsing bug, where it would accept leading zeroes like `RX01` and `S00`

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -36,6 +36,7 @@
 //! ```
 
 #![deny(clippy::all, unused_must_use, unused_imports)]
+#![feature(or_patterns)]
 
 // Re-export dependencies that consumers may need
 pub use validator;
--- a/core/src/parse.rs
+++ b/core/src/parse.rs
@@ -50,7 +50,17 @@ trait Parse<'a>: Sized {
 // covers StackId and UserRegisterId
 impl<'a> Parse<'a> for usize {
     fn parse(input: RawSpan<'a>) -> ParseResult<'a, Self> {
-        map_res(digit1, |s: RawSpan| s.fragment().parse::<usize>())(input)
+        map_res(digit1, |s: RawSpan| {
+            let frag = s.fragment();
+
+            // If the string has unnecessary leading zeroes, reject it. Use an
+            // empty error for convenience, its value won't be used anyway.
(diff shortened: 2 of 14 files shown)


core/src/lib.rs
```diff
@@ -36,6 +36,7 @@
 //! ```
 
 #![deny(clippy::all, unused_must_use, unused_imports)]
+#![feature(or_patterns)]
```
LucasPickering, line 39: this isn't really necessary but it's nice. I checked the rustfmt features that we're still on nightly for, and it doesn't seem like  they're getting stablized any time soon. If we want to switch to stable at some point, it wouldn't be hard to remove this feature.

core/src/parse.rs
```diff
@@ -50,7 +50,17 @@ trait Parse<'a>: Sized {
...
+            let frag = s.fragment();
+
+            // If the string has unnecessary leading zeroes, reject it. Use an
+            // empty error for convenience, its value won't be used anyway.
+            if frag.len() > 1 && frag.chars().next() == Some('0') {
+                Err(())
+            } else {
+                s.fragment().parse::<usize>().map_err(|_| ())
+            }
+        })(input)
```
LucasPickering, line 63: This is all the logic for the parsing bug fix. The only other code in here related to the bug fix is the tests.

core/src/validate.rs
```diff
@@ -177,8 +178,10 @@ fn validate_writable(
 ) {
     // Only User registers are writable, all others cause an error.
     match reg_ref_node {
-        Node(RegisterRef::User(_), _) => {}
-        Node(_, span) => errors.push((CompileError::UnwritableRegister, *span)),
+        Node(RegisterRef::Null | RegisterRef::User(_), _) => {}
+        Node(RegisterRef::InputLength | RegisterRef::StackLength(_), span) => {
```
LucasPickering, line 182: This is the new `or_patterns` feature. It lets you use the `|` operator deep inside patterns, instead of just at the top level.

PR #79, opened 2020-05-24
Add tests for wasm functions

LucasPickering:
I did find one bug, which was the instructions which generate a runtime error doun't count as a CPU cycle. I fixed that.


core/src/util.rs
```diff
@@ -218,7 +218,6 @@ macro_rules! debug {
 }
 
 // Only needed in tests
-#[cfg(test)]
```
LucasPickering, line 221: `cfg(test)` doesn't get set frfom outside the crate, and we need this impl in the wasm crate tests too now.

PR #81, opened 2020-05-28
Start adding OpenID

JRMurr:
WARNING SPAGHETTI 

Inital stab at adding OpenID. Just creating users for now with no logout (soon ™️). `actix-identity` handles adding a cookie to the frontend session with their login token so we should be able to add that to graphql requests to auth. 

I basically copy pastad from https://docs.rs/openid/0.3.1/openid/


api/src/server/auth.rs
```diff
@@ -0,0 +1,140 @@
...
+        .await
+    {
+        Ok(res) => res,
+        Err(e) => panic!(e),
+    }
+}
+
+// The frontend will redirect to this before being sent off to the
+// actual openid provdider
+// TODO: add query param to say which provider to use
```
LucasPickering, line 40: Can you make this a doc comment instead?

api/src/server/auth.rs
```diff
@@ -0,0 +1,140 @@
...
+        .finish()
+}
+
+#[derive(Deserialize, Debug)]
+pub struct LoginQuery {
+    code: String,
+}
+
+// Exchanges the access token from the inital login in the openid provider
+// for a normal token
```
LucasPickering, line 61: same here

api/src/server/auth.rs
```diff
@@ -0,0 +1,140 @@
...
+        .map_err(ResponseError::from)?;
+
+    Ok(Some((token, userinfo)))
+}
+
+fn host(path: &str) -> String {
+    "http://localhost:8000".to_string() + path
+}
+
+// Provider redirects back to this route after the login
```
LucasPickering, line 93: and here

api/src/server/auth.rs
LucasPickering, line 40: Not a huge deal, but I think it'd be better to use a route param, like `/blah/<provider>`

api/src/server/auth.rs
```diff
@@ -0,0 +1,140 @@
...
+    {
+        Ok(res) => res,
+        Err(e) => panic!(e),
+    }
+}
+
+// The frontend will redirect to this before being sent off to the
+// actual openid provdider
+// TODO: add query param to say which provider to use
+#[get("/oauth2/authorization/oidc")]
```
LucasPickering, line 41: all api routes should start with `/api/`. Also, is this route layout a convention? It seems way longer than it needs to be

api/src/server/auth.rs
```diff
@@ -0,0 +1,140 @@
...
+use actix_identity::Identity;
+use actix_web::{get, http, web, HttpResponse, Responder};
+use diesel::{prelude::RunQueryDsl, PgConnection};
+use openid::{Client, DiscoveredClient, Options, Token, Userinfo};
+use reqwest::Url;
+use serde::Deserialize;
+use std::{collections::HashMap, sync::RwLock};
+pub struct Sessions {
+    pub map: HashMap<String, (User, Token, Userinfo)>,
+}
```
LucasPickering, line 16: Not sure if you're planning this, but this should definitely be a DB table, so the locking is much more efficient and it's persistent through restarts etc.

api/src/server/auth.rs
```diff
@@ -0,0 +1,140 @@
+use crate::{
```
LucasPickering, line 1: Can you just generally add some more comments to the bodies of these functions? The logic isn't 100% clear.

api/src/server/auth.rs
```diff
@@ -0,0 +1,140 @@
...
+            identity.remember(id.clone());
+            sessions
+                .write()
+                .unwrap()
+                .map
+                .insert(id, (user, token, userinfo));
+            // TODO: add redirect path to state param so we dont always redirect
+            // to the homepage
+            Ok(HttpResponse::Found()
+                .header(http::header::LOCATION, "http://localhost:3000/")
```
LucasPickering, line 126: This should redirect to just `/`. If you add the `/api/` to the start of the routes, then you can access them through the proxy on 3000, so that you can redirect to FE routes without needing the hostname.

api/src/server/auth.rs
```diff
@@ -0,0 +1,140 @@
...
+        Ok(None) => {
+            eprintln!("login error in call: no id_token found");
+
+            Ok(HttpResponse::Unauthorized().finish())
+        }
+        Err(err) => {
+            eprintln!("login error in call: {:?}", err);
+
+            Ok(HttpResponse::Unauthorized().finish())
+        }
```
LucasPickering, line 138: Since these are client errors, I don't think there's any need to log the errors. Also for future reference, you can use the macros from [the log crate](https://docs.rs/log/0.4.8/log/) for better logging.

LucasPickering: oh also, can you add some comments explaining when these error cases would be hit? And since they return the same value, you can collapse them into one match branch.

api/src/server/mod.rs
```diff
@@ -42,17 +49,36 @@ pub async fn run_server(pool: Pool, host: String) -> io::Result<()> {
     // Init GraphQL schema
     let gql_schema = Arc::new(create_gql_schema());
 
+    let google_client_id = env::var("OPENID_CLIENT_ID").unwrap();
+    let google_client_secret = env::var("OPENID_CLIENT_SECRET").unwrap();
+    let google_url = "https://accounts.google.com";
+    let client =
+        make_client(google_client_id, google_client_secret, google_url).await;
```
LucasPickering, line 56: Do you have a plan for how to handle multiple clients? Should we store them in a map of `name:client`? Also, it'd be nice to move as much of the auth logic out of this file and into `auth.rs` as possible. Maybe once we have multiple clients we could offload all that logic into that file?

Also, does this make a during call during the `.await`? I don't think we should be blocking server start on an outside network call. Is there a way around that?

api/src/server/mod.rs
```diff
@@ -42,17 +49,36 @@ pub async fn run_server(pool: Pool, host: String) -> io::Result<()> {
     // Init GraphQL schema
     let gql_schema = Arc::new(create_gql_schema());
 
+    let google_client_id = env::var("OPENID_CLIENT_ID").unwrap();
+    let google_client_secret = env::var("OPENID_CLIENT_SECRET").unwrap();
```
LucasPickering, line 53: Should these be named `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` instead?

docker-compose.yml
```diff
@@ -31,6 +31,8 @@ services:
       RUST_BACKTRACE: 1
       SERVER_HOST: 0.0.0.0:8000
       RUST_LOG: info
+      OPENID_CLIENT_ID: ${OPENID_CLIENT_ID}
+      OPENID_CLIENT_SECRET: ${OPENID_CLIENT_SECRET}
```
LucasPickering, line 35: I think if you just don't pass a value, it'll automatically forward whatever value is in the env.

frontend/src/components/LoginPage.tsx
```diff
@@ -0,0 +1,24 @@
...
+const LoginPage: React.FC = () => {
+  return (
+    <Grid container justify="center">
+      <Grid item md={4} sm={8} xs={12}>
+        <Button
+          color="primary"
+          // variant="outlined"
+          component={Link}
+          // TODO: good url
+          to="http://localhost:8000/oauth2/authorization/oidc"
```
LucasPickering, line 15: Once you add `/api/` to the routes you should be able to do this now.

api/src/server/auth.rs
JRMurr: Yea ill do that in the follow up

api/src/server/auth.rs
JRMurr: Not sure if its a convention tbh, just went off the example for the lib. Ill rename it

api/src/server/mod.rs
JRMurr: Yea i was thinking a map of name to client. We could probably do all the init stuff in `auth.rs` and just return that map here so this is less bloated.

> Also, does this make a during call during the .await?
Yea, the ` match DiscoveredClient::discover` in `make_client` calls out to a standard url that returns info about how the provider is configured. We might be able to set up a wrapper around the openid client that says if it is ready or not. Then on the first login attempt or w/e we call out to the discover url to configure it. But really it shouldn't take too long on startup to init. Im done for either

docker-compose.yml
JRMurr: like this?
```
OPENID_CLIENT_ID: 
OPENID_CLIENT_SECRET: 
```

It got set to null if i did that

api/src/server/auth.rs
JRMurr: It might be a little weird since the provider redirects to `localhost:8000/api/callback` so idk if the proxy will pick that up, ill try though

docker-compose.yml
JRMurr: wait nvm, my `.env` file got wiped when i switched branches lol

api/src/server/auth.rs
JRMurr: yea when you do `/` it just goes to `localhost:8000/`

api/src/server/mod.rs
LucasPickering: ^

.env.example
```diff
@@ -0,0 +1,3 @@
+DOCKER_TAG=latest
+OPENID_CLIENT_ID=<ID>
+OPENID_CLIENT_SECRET=<SECRET>
```
LucasPickering, line 3: So env files with docker-compose are kinda jank, but I _think_ we can leave the existing `.env` file, and use a separate one for the secrets. Can you try putting the original `.env` back with just the `DOCKER_TAG`, then add a file like `api/vars.env`, and under the api container in the `docker-compose.yml`, you set `env_file: api/vars.env`? I think it'll load both and we can add `api/vars.env` to the gitignore. Then have like `api/vars.example.env` with placeholder/empty values?

api/src/error.rs
```diff
@@ -44,6 +44,14 @@ pub enum ResponseError {
...
     #[fail(display = "Database error: {}", 0)]
     DieselError(#[cause] diesel::result::Error),
+
+    /// Wrapper for openid's client error type
+    #[fail(display = "{}", 0)]
+    OpenIdClientError(#[cause] ClientError),
+
+    /// Wrapper for openid's error type
+    #[fail(display = "{}", 0)]
+    OpenIdError(#[cause] OpenIdError),
```
LucasPickering, line 54: Not sure if this is on your branch since I did it recently, but on master these error types are broken up by client vs server. Could you move these around in the enum to fit?

api/src/server/auth.rs
```diff
@@ -0,0 +1,144 @@
...
+    pub map: HashMap<String, (User, Token, Userinfo)>,
+}
+/// Setup a client for a given provider
+/// based on the id, secret, and provider url
+pub async fn make_client(
+    client_id: String,
+    client_secret: String,
+    issuer_url: &str,
+) -> Client {
+    let redirect = Some(host("/api/oidc/callback"));
```
LucasPickering, line 25: Why does the client need to know the redirect in order to be created? Doesn't that only need to be known by the provider? Since the func takes an `Option`, what happens if you just pass `None`?

api/src/server/mod.rs
JRMurr: O I did that originally but the linter was sad at me, I can remove that rule but this will probably be removed when we support multiple providers so idk if its worth

api/src/server/auth.rs
```diff
@@ -87,11 +92,11 @@ async fn request_token(
...
 
 fn host(path: &str) -> String {
-    "http://localhost:8000".to_string() + path
+    "http://localhost:3000".to_string() + path
 }
 
-// Provider redirects back to this route after the login
-#[get("/login/oauth2/code/oidc")]
+/// Provider redirects back to this route after the login
+#[get("/api/oidc/callback")]
```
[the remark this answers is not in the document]
JRMurr: We could do a route param or put the provider in the state query param, which ever we want

api/src/server/auth.rs
JRMurr: Since you can specify multiple valid redirects in the provider you need to tell it which one to use when you initally send the user to the provider. Not sure why the lib made it an option type tbh

api/src/error.rs
JRMurr: How should they be broken up? I think client error means the openid client messed up/ the provider messed up, Same with `OpenIdError`. Tbh i cant tell the diff btwn them

.env.example
JRMurr: Now that i think of it, we probaly should just have a yaml/json file that is a map like
```yaml
<provider name>:
       client_id: <id>
       client_secret: <secret>
       issuer_url: <url>
```

That way we dont need to make new env vars for each provider

PR #86, opened 2020-06-03
Add GQL operations to create and mutate hardware and program specs (spec ops)

LucasPickering:
- Added the new `name` column to the hardware spec accessor
- Added the new `name` and `description` columnds to the program spec accessor
- Added for new mutations
  - `createHardwareSpec`
  - `updateHardwareSpec`
  - `createProgramSpec`
  - `updateProgramSpec`

For the create mutations, I made the slugs auto-generated based on the name. For the update mutations, you can change the name but the slug doesn't get re-generated. This is to prevent slugs getting invalidated and breaking links.

Then some other random shit:
- Added a `db-reset` task to `cargo make` for the API crate
- Fixed up some missing DB constraints
- Split all the mutations into a separate file
- Added the new `NoUpdate` API error and a field in `DbErrorConverter` to generate it
- Split apart the `DbErrorConverter` tests into multiple funcs

### DIFF
--- a/api/migrations/2020-05-30-210421_init_models/up.sql
+++ b/api/migrations/2020-05-30-210421_init_models/up.sql
@@ -1,7 +1,7 @@
 CREATE TABLE hardware_specs (
     id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
     slug VARCHAR(50) NOT NULL UNIQUE CHECK (char_length(slug) > 0),
-    name VARCHAR(50) NOT NULL UNIQUE CHECK (char_length(slug) > 0),
+    name VARCHAR(50) NOT NULL UNIQUE CHECK (char_length(name) > 0),
     num_registers INTEGER NOT NULL CHECK(num_registers >= 1),
     num_stacks INTEGER NOT NULL CHECK(num_stacks >= 0),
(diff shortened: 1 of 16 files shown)


api/migrations/2020-05-30-210421_init_models/up.sql
```diff
@@ -16,24 +16,26 @@ CREATE TABLE program_specs (
     name VARCHAR(50) NOT NULL CHECK (char_length(name) > 0),
     description TEXT NOT NULL,
     hardware_spec_id UUID NOT NULL REFERENCES hardware_specs(id),
-    input SMALLINT[] NOT NULL,
-    expected_output SMALLINT[] NOT NULL,
-    UNIQUE(slug, hardware_spec_id)
+    -- TODO add constraint to make sure all values in range
+    input INT[] NOT NULL CHECK (array_length(input, 1) <= 256),
+    expected_output INT[] NOT NULL CHECK (array_length(input, 1) <= 256),
```
LucasPickering, line 21: I changed these from 16-bit to 32-bit. 16-bit is what the actual GDLK values are, but it's easier if the DB types match what comes in from the API, then just apply the proper range changes with constraints. Also, a signed 16-bit int can't actually hold an unsigned 16-bit so shit was a lil fucked.

PR #89, opened 2020-06-07
Add parse error for empty program

LucasPickering:
You were right. It doesn't really make sense to compile an empty program. Now, with an empty file, it shows a compile error in the IDE. It seems a _little_ weird, but I think it's fine cause the program can't compile so it makes sense to show the user why.


core/src/parse.rs
```diff
@@ -442,10 +463,46 @@ mod tests {
...
+        assert_eq!(
+            parse("  \n  LBL:").unwrap().body,
+            vec![Node(
+                Statement::Label(Node(
+                    LabelDecl("LBL".into()),
+                    span(2, 3, 2, 7)
+                )),
+                span(2, 3, 2, 7)
+            )]
+        );
```
LucasPickering, line 505: Testing errors here is a bitch, so I just added a label so that these are still passing.

PR #91, opened 2020-06-13
Add more docs content, fix hash links

LucasPickering:
- Added more content to the docs page
- Cleaned up some link stuff. We shouldn't have to import any links from other packages now, just use the local `Link` or `UnstyledLink`
- Fixed links with hashes, so we can link between different parts of one page now


frontend/.eslintrc.js
```diff
@@ -8,13 +8,20 @@ module.exports = {
...
     project: './tsconfig.json',
   },
-  plugins: ['@typescript-eslint', 'react', 'react-hooks', 'prettier'],
+  plugins: [
+    '@typescript-eslint',
+    'react',
+    'react-hooks',
+    'prettier',
+    'jsx-a11y',
+  ],
```
LucasPickering, line 17: I added a new eslint plugin for accessibility rules

frontend/package.json
```diff
@@ -15,53 +15,55 @@
     ]
   },
   "dependencies": {
-    "@material-ui/core": "^4.9.11",
```
LucasPickering, line 18: I upgraded all our dependencies

frontend/src/components/docs/InstructionDocs.tsx
```diff
@@ -0,0 +1,246 @@
...
+  },
+  {
+    name: 'ADD',
+    summary: 'Add a value to a register.',
+    moreInfo: (
+      <>
+        The result is stored in the register. See{' '}
+        <Link to="#values--overflow-and-underflow">
+          arithmetic overflow and underflow
+        </Link>
```
JRMurr, line 75: This is mostly me thinking out load and not a big deal but you think we could make it fill in the link text dynamically based on the id we link to? That way if we want to change the title or w/e we wouldn't need to update all the links

PR #95, opened 2020-06-19
More openid stuff

JRMurr:
Setup the `user_providers` table to map the oidc stuff to a given user. We still auto set the username but should be better soon™️

### DIFF
--- a/api/config/default.json
+++ b/api/config/default.json
@@ -1,7 +1,7 @@
 {
   "server_host": "0.0.0.0:8000",
   "open_id": {
-    "host_url": "localhost:3000",
+    "host_url": "http://localhost:3000",
     "providers": {
       "google": {
         "issuer_url": "https://accounts.google.com"
--- a/api/src/server/auth.rs
+++ b/api/src/server/auth.rs
@@ -1,14 +1,15 @@
 use crate::{
     config::{OpenIdConfig, ProviderConfig},
     error::ResponseError,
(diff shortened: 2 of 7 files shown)


api/config/default.json
```diff
@@ -1,7 +1,7 @@
 {
   "server_host": "0.0.0.0:8000",
   "open_id": {
-    "host_url": "localhost:3000",
+    "host_url": "http://localhost:3000",
```
LucasPickering, line 4: oh shit maybe that's why i couldn't get it to work...

api/src/server/auth.rs
```diff
@@ -141,32 +219,67 @@ pub async fn route_login(
...
+            match existing_user_provider {
+                Some(user_provider) => match user_provider.user_id {
+                    Some(user_id) => {
+                        // User already exists so normal login
+                        let user: User = users::table
+                            .filter(users::dsl::id.eq(user_id))
+                            .first(conn)
+                            .optional()
+                            .map_err(ResponseError::from)?
+                            .unwrap();
```
LucasPickering, line 238: The `optional` here is completely unnecessary, since you're just unwrapping the option anyway. You also can use `get_result` instead of `first`, so that it errors if there's more than one user (which obv is impossible but best practice is to use `get_result` whenever you expect exactly 1). So
`let user: User = users::table.filter(...).get_result().map_err(...)?;`

PR #97, opened 2020-06-24
Add date columns to userProgram, store source code in browser local storage

LucasPickering:
The main purpose here is to store the user's source code in the browser's local storage, so that they don't lose unsaved changes (and this would make it easier to have the IDE work entirely offline). To make this work tho, I added two new columns to `user_programs`:

- `created` (this one wasn't exactly necessary but it seemed useful and it was easy)
- `last_modified`

I also added a prompt that appears if you navigate away from the page without saving your changes. Unfortunately it only works if you're staying within the app, so hitting refresh or the back button won't trigger it. We still have local storage to cover those cases tho.

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -14,12 +14,13 @@ actix-web = "2.0"
 actix-web-actors = "2.0"
 actix-rt = "1.0"
 base64 = "^0.12.0"
+chrono = "0.4"
 config = { version = "0.10", default-features = false, features = ["json"] }
-diesel = { version = "^1.4.3", features = ["postgres", "r2d2", "uuidv07"] }
+diesel = { version = "^1.4.3", default-features = false, features = ["chrono", "postgres", "r2d2", "uuidv07"] }
 env_logger = "0.7"
 failure = "0.1"
 gdlk = { path = "../core" }
-juniper = "0.14.2"
+juniper = { version = "0.14.2", default-features = false, features = ["chrono"] }
 # juniper-from-schema = "0.5.2"
 juniper-from-schema = { git = "https://github.com/LucasPickering/juniper-from-schema", branch = "fragment-bug-workaround" }
 log = "^0.4.8"
--- a/frontend/src/components/ide/ProgramIde.tsx
+++ b/frontend/src/components/ide/ProgramIde.tsx
@@ -21,6 +21,7 @@ import { HardwareSpec, ProgramSpec } from 'gdlk_wasm';
 import useDebouncedValue from 'hooks/useDebouncedValue';
 import { assertIsDefined } from 'util/guards';
 import NotFoundPage from 'components/NotFoundPage';
+import { Prompt } from 'react-router-dom';
 
 const useLocalStyles = makeStyles(({ palette, spacing }) => {
   const border = `2px solid ${palette.divider}`;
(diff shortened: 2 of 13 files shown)


api/Cargo.toml
```diff
@@ -14,12 +14,13 @@ actix-web = "2.0"
...
 base64 = "^0.12.0"
+chrono = "0.4"
 config = { version = "0.10", default-features = false, features = ["json"] }
-diesel = { version = "^1.4.3", features = ["postgres", "r2d2", "uuidv07"] }
+diesel = { version = "^1.4.3", default-features = false, features = ["chrono", "postgres", "r2d2", "uuidv07"] }
 env_logger = "0.7"
 failure = "0.1"
 gdlk = { path = "../core" }
-juniper = "0.14.2"
+juniper = { version = "0.14.2", default-features = false, features = ["chrono"] }
```
LucasPickering, line 23: Rust's stdlib doesn't have date support, `chrono` is the most popular lib for it

frontend/src/components/ide/ProgramIde.tsx
```diff
@@ -217,6 +242,10 @@ const ProgramIde: React.FC<{
         />
         <StackInfo className={localClasses.stackInfo} />
         <CodeEditor className={localClasses.editor} />
+        <Prompt
+          when={sourceCode !== userProgram.sourceCode}
+          message="You have unsaved changes. Are you sure you want to leave?"
+        />
```
LucasPickering, line 248: This is all the logic for the unsaved changes prompt. React router handles the rest.

frontend/src/util/storage.ts
```diff
@@ -0,0 +1,54 @@
...
+    const storedObj: StorageValue<T> = {
+      value,
+      lastModified: new Date(), // This will be serialized as an ISO string
+    };
+    window.localStorage.setItem(this.key, JSON.stringify(storedObj));
+  }
+
+  public delete(): void {
+    window.localStorage.removeItem(this.key);
+  }
```
LucasPickering, line 53: Right now we're never deleting these keys, but I left this in anyway. I'm trying to think if there's a good way to clean up the storage automatically, since it could get quite large with one entry per user_program. We could have it delete whenever you leave the editor, if all your changes are saved. But then it's not available later if you wanna edit offline so idk if that's the best solution.

frontend/src/components/ide/ProgramIde.tsx
LucasPickering, line 248: This is the only logic needed for the prompt-on-exit. React-router handles the rest.

frontend/src/hooks/useStaticValue.ts
```diff
@@ -0,0 +1,24 @@
...
+    initialized: false,
+    value: undefined,
+  });
+
+  if (!ref.current.initialized) {
+    ref.current = { initialized: true, value: init() };
+  }
+
+  // At this point we know `value` is a T, but we  have to convince the typechecker
+  return ref.current.value as T;
```
JRMurr, line 21: Its probably overkill to use this but did you lookinto a type guard so we dont need the cast https://www.typescriptlang.org/docs/handbook/advanced-types.html#using-type-predicates

LucasPickering: Yeah so I thought about that, but because `T` is completely unknown, the caller would have to supply a type guard that can handle type `T`, which would be real annoying. Since we know for 100% sure that `ref.current.value` is a `T` at this point (because we _always_ assign it before hitting this for the first time, and we know for sure that `init()` returns a `T`, I think the cast is safe.

LucasPickering:
@JRMurr I did some googling and it looks like we could use the standard library's [onbeforeunload](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload) to handle navigation outside the app. I can look into that in a separate branch, don't wanna bloat this one anymore.

PR #98, opened 2020-06-25
Remove sessions, and clean up some auth code

LucasPickering:
Removed the `Sessions` type and everything that used it, since we decided we don't need it. I also tried to clean up some of the auth code by doing more error handling, and trying to reduce how much data we need to pass between functions.

### DIFF
--- a/api/src/server/auth.rs
+++ b/api/src/server/auth.rs
@@ -14,12 +14,7 @@ use diesel::{
 use openid::{Client, DiscoveredClient, Options, Token, Userinfo};
 use reqwest::Url;
 use serde::Deserialize;
-use std::{collections::HashMap, sync::RwLock};
-
-// TODO: make this a db table so its persistent through restarts
-pub struct Sessions {
-    pub map: HashMap<String, (User, Token, Userinfo)>,
-}
+use std::collections::HashMap;
 
 /// Map of provider name to configured [Client]
 pub struct ClientMap {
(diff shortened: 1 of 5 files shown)


api/src/server/auth.rs
```diff
@@ -96,110 +91,71 @@ pub struct LoginQuery {
...
-        .email
-        .clone()
-        .unwrap()
-        .chars()
-        .take(20)
-        .collect::<String>();
-    let user = make_user(&user_provider, &username, conn);
-    login_user(&user_provider, user, token, userinfo, identity, sessions)
+    Ok(())
 }
```
JRMurr, line 143: So the main reason i had the two funcs was because `init_user` was gonna be called when we wanted to redirect to the username setup and `make_user` would handle the call after the redirect if that makes sense.

api/src/server/auth.rs
```diff
@@ -208,81 +164,50 @@ pub async fn route_login(
...
-                },
-                None => {
-                    // user_provider not found (first login) so make the row
-                    // then init the user
-                    let user_provider: UserProvider = NewUserProvider {
-                        sub: &sub,
-                        provider_name: &provider_name,
-                    }
+    // Not sure when this can be None, hopefully never??
+    let sub: &str = userinfo.sub.as_ref().unwrap();
```
JRMurr, line 178: It *should* never be None but i hope it would error before that in `request_token` if is not there

PR #99, opened 2020-06-27
Prompt for unsaved changes in IDE

LucasPickering:
Added a second IDE prompt, which gets triggered for browser navs (which react-router can't trap). I moved this functionality into a utility component to keep it clean and in case we need to use it again. I tested in Firefox and Chrome, lookin good.

### DIFF
--- a/frontend/src/components/ide/ProgramIde.tsx
+++ b/frontend/src/components/ide/ProgramIde.tsx
@@ -1,10 +1,4 @@
-import React, {
-  useState,
-  useEffect,
-  useMemo,
-  useCallback,
-  useRef,
-} from 'react';
+import React, { useState, useEffect, useCallback, useRef } from 'react';
 import { RelayProp, createFragmentContainer } from 'react-relay';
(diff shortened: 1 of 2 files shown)


frontend/src/components/ide/ProgramIde.tsx
```diff
@@ -90,25 +84,23 @@ const ProgramIde: React.FC<{
...
-  const wasmProgramSpec = useMemo(
+  const wasmProgramSpec = useStaticValue(
     () =>
       new ProgramSpec(
         Int16Array.from(programSpec.input),
         Int16Array.from(programSpec.expectedOutput)
-      ),
-    [programSpec]
+      )
   );
```
LucasPickering, line 104: I also realized I could use our new `useStaticValue` hook here too. The React team have said that in the future, they may change `useMemo` to call the memoized function more than necessary if it's more performant, which in our case might cause some fuqqery. Have it be guaranteed stable should be a bit safer.

PR #101, opened 2020-07-03
Use authenticated user in graphql responses

LucasPickering:
This adds the logged-in user to the graphql context, and updates all the user program API handlers to reference that user instead of the `tmp_user` hack. I had to add a bunch of tests to account for new edge cases that occur when the user isn't logged in, or tries to access another user's programs.

Currently this breaks some shit on the frontend if you're not logged in (cause it tries to fetch the user programs but that gives an error). I removed one of these cases but the others i left in place. I'm gonna work on that soon.

### DIFF
--- a/api/src/server/gql/mutation.rs
+++ b/api/src/server/gql/mutation.rs
@@ -1,7 +1,7 @@
 use crate::{
     error::{DbErrorConverter, ResponseResult},
     models,
-    schema::{hardware_specs, program_specs, user_programs, users},
+    schema::{hardware_specs, program_specs, user_programs},
     server::gql::{
         Context, CreateHardwareSpecInput, CreateHardwareSpecPayload,
         CreateProgramSpecInput, CreateProgramSpecPayload,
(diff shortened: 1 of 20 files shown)


api/src/server/gql/mutation.rs
```diff
@@ -239,14 +240,19 @@ impl MutationFields for Mutation {
...
-                .optional();
+        // Update the row, returning the new value. If the row doesn't
+        // exist, this will return None.
+        let result: Result<Option<models::UserProgram>, _> = diesel::update(
+            user_programs::table
+                .find(modified_user_program.id)
+                // Extra filter to make sure the requesting user is the owner of
+                // the row.
+                .filter(user_programs::columns::user_id.eq(user.id)),
+        )
```
JRMurr, line 251: Should we throw an error saying you dont have access to this program instead of doing nothing?

LucasPickering: I think in general, if a user doesn't have read perms on an object, then any time they try to do something to it we should pretend like it doesn't exist, to prevent leaking any data. In this case it wouldn't be a big deal cause we would just be leaking that a certain guid exists in the DB, but i think it's still the best practice.

LucasPickering:
@JRMurr There were some import conflicts on this after merging the other PR so I fixed those, didn't make any changes otherwise tho.

api/src/server/gql/mutation.rs
JRMurr: image

PR #102, opened 2020-07-03
Improve cookie security

LucasPickering:
Made some cookie policy changes:
- Changed `Max-Age` to 24h (it was just `Session` before, so in a way this is less secure but I think 24h is still fine)
- Set `Secure: true`
  - This obv requires https to work, so I enabled it in dev. create-react-app has this built in, it just creates a self-signed cert. You'll have to use `https://localhost:3000` from now on
- Used the actual secret key for encryption (all zeroes isn't exactly secure)
- Set `SameSite: Lax`, which prevents CSRF attacks

I also fixed a random UI bug, where the IDE wouldn't show compilation results on the first render.

### DIFF
--- a/api/config/default.json
+++ b/api/config/default.json
@@ -1,7 +1,7 @@
 {
   "server_host": "0.0.0.0:8000",
   "open_id": {
-    "host_url": "http://localhost:3000",
+    "host_url": "https://localhost:3000",
     "providers": {
       "google": {
         "issuer_url": "https://accounts.google.com"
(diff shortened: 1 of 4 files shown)


api/config/default.json
```diff
@@ -1,7 +1,7 @@
 {
   "server_host": "0.0.0.0:8000",
   "open_id": {
-    "host_url": "http://localhost:3000",
+    "host_url": "https://localhost:3000",
     "providers": {
       "google": {
         "issuer_url": "https://accounts.google.com"
```
JRMurr, line 7: this is really minor and doesnt need to be addressed in this pr but should we move `host_url` to the root of the config?

LucasPickering: I was actually thinking the same thing. Right now we're just using it for OpenID stuff so I think it's fine where it is but if we start using it for anything else we should def move it out. I'm fine either way tho

PR #105, opened 2020-07-06
Add initializeUser mutation

LucasPickering:
Adds a new mutation called `initializeUser` that lets you set your username for the first time. I thought about just making this a generic `setUsername` mutation so that we can let people change their usernames in the future, but I figured at some point we may need extra data in `initializeUser` beyond just the username. We can add a separate `setUsername` later, and the logic for it should be simpler anyway.

### DIFF
--- a/api/src/models/user.rs
+++ b/api/src/models/user.rs
@@ -4,6 +4,7 @@ use diesel::{
     sql_types::Text, Identifiable, Queryable,
 };
 use uuid::Uuid;
+use validator::Validate;
 
 /// Expression to filter users by username
 pub type WithUsername<'a> =
@@ -30,9 +31,10 @@ impl User {
     }
 }
 
-#[derive(Debug, Default, PartialEq, Insertable)]
+#[derive(Debug, Default, PartialEq, Insertable, Validate)]
 #[table_name = "users"]
 pub struct NewUser<'a> {
+    #[validate(length(min = 1))]
     pub username: &'a str,
 }
 
--- a/api/src/models/user.rs
+++ b/api/src/models/user.rs
@@ -34,7 +34,7 @@ impl User {
 #[derive(Debug, Default, PartialEq, Insertable, Validate)]
 #[table_name = "users"]
 pub struct NewUser<'a> {
-    #[validate(length(min = 1))]
+    #[validate(length(min = 1, max = 20))]
     pub username: &'a str,
 }
 
--- a/api/src/server/gql/mutation.rs
+++ b/api/src/server/gql/mutation.rs
@@ -1,21 +1,24 @@
 use crate::{
-    error::{DbErrorConverter, ResponseResult},
+    error::{DbErrorConverter, ResponseError, ResponseResult},
     models,
-    schema::{hardware_specs, program_specs, user_programs},
+    schema::{
+        hardware_specs, program_specs, user_programs, user_providers, users,
+    },
     server::gql::{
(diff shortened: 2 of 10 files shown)


api/src/server/gql/mutation.rs
```diff
@@ -25,6 +28,73 @@ use validator::Validate;
...
+                                .eq(Some(created_user.id)),
+                        )
+                        .execute(conn)?;
+
+                        if updated_rows == 0 {
+                            Err(ResponseError::NotFound)
+                        } else {
+                            Ok(created_user)
+                        }
+                    })?;
```
LucasPickering, line 89: Can you make sure this transaction logic makes sense? I doubled checked in the diesel source, and if this lambda returns any `Err` it will roll back. I also tried to verify that in the test. We should be good here if there are simultaneous requests, right? One should go through properly while the other reverts.

api/tests/test_initialize_user.rs
```diff
@@ -0,0 +1,197 @@
+#![deny(clippy::all)]
```
LucasPickering, line 1: Could you read through these tests cases and see if you think we're missing any? I think I got all the edges cases

api/src/server/gql/mutation.rs
JRMurr: > We should be good here if there are simultaneous requests, right? 

What would be simultaneous, isn't it synchronous since it updates the `users` table then `user_providers`?

api/src/models/user.rs
```diff
@@ -30,9 +31,10 @@ impl User {
     }
 }
 
-#[derive(Debug, Default, PartialEq, Insertable)]
+#[derive(Debug, Default, PartialEq, Insertable, Validate)]
 #[table_name = "users"]
 pub struct NewUser<'a> {
+    #[validate(length(min = 1))]
```
JRMurr, line 37: Should this also have the max len of 20 validate or should we just make username be `TEXT` instead of varchar?

api/tests/test_initialize_user.rs
```diff
@@ -0,0 +1,204 @@
...
+                "message": "This resource already exists",
+                "locations": [{"line": 5, "column": 9}],
+                "path": ["initializeUser"],
+            })]
+        )
+    );
+}
+
+/// Setting a username that doesn't pass validation should return an error
+#[test]
```
JRMurr, line 124: If we decide to keep a max username len we should add a test for that but otherwise these seem to cover everything i can think of

api/src/models/user.rs
LucasPickering: That's a good point. I'm still planning on switch to just using the DB constraints for validation at some point, but for now since we're still using `validator` I'll add that in

api/src/server/gql/mutation.rs
LucasPickering: If someone sent the same mutation request twice simultaneously, we could potentially have multiple threads in this critical section at the same time.

api/tests/test_initialize_user.rs
LucasPickering: :+1:

api/src/server/gql/mutation.rs
JRMurr: I think it would error out correctly on commit in that case but tbh im not sure. We can just leave it as a thing to look into since i doubt it would ever actually happen

PR #106, opened 2020-07-10
Add user data to UI and other user-related stuff

LucasPickering:
- Moved log in/out buttons to top-right
  - I also got rid of the login page in favor of just a dropdown menu
- Loaded user status into the UI so now we're storing it in global read-only state
- Which button we show is now contextual based on whether user is logged in or not
- Hide the User Program data on the hardware spec and program spec pages if the user isn't logged in
- Used the `state` param to remember the user's previous location after login


api/src/server/auth.rs
```diff
@@ -70,13 +85,18 @@ pub async fn build_client_map(open_id_config: &OpenIdConfig) -> ClientMap {
...
-    let oidc_client = client_map.get_client(&provider_name)?;
+    let provider_name: &str = &params.0;
+    let oidc_client = client_map.get_client(provider_name)?;
+    let state = AuthState {
+        next: query.next.as_deref(),
+    };
 
     let auth_url = oidc_client.auth_url(&Options {
         scope: Some("email".into()),
+        // Serialization shouldn't ever fail so yeet that shit outta the Result
```
JRMurr, line 98: This technically could fail if they just hit this endpoint manually but idk if that matters. Does `actix_web` kill the server on panic?

JRMurr: I also ran into issues before where some (bad) openid providers would mess with the formatting of the state paraem but idk if we should care about that either

frontend/src/components/user/LoginMenu.tsx
```diff
@@ -0,0 +1,103 @@
...
+              vertical: 'top',
+              horizontal: 'right',
+            }}
+            open={open}
+            onClose={() => setOpen(false)}
+          >
+            <MenuItem
+              onClick={async () => {
+                await fetch('/api/logout', { method: 'POST' });
+                // fuckin yeet em back to the home page yeehaw
```
JRMurr, line 53: image

api/src/server/auth.rs
LucasPickering: Here's what the docs say about when `serde_json::to_string` can fail:

> Serialization can fail if T's implementation of Serialize decides to fail, or if T contains a map with non-string keys.

Since `T` is `AuthState` which is just a simple derived serializer, it can never fail. And we don't have any maps in here, so the second part can't happen either. So `serde_json::to_string` can literally never fail in this case. But yeah if there is a panic, actix kills that thread (but i think the rest of the server keeps running).

PR #107, opened 2020-07-11
Add UI for setting the username

LucasPickering:
Now, after you log in, it automatically pops up a modal to ask for a new username.

### DIFF
--- a/api/src/server/auth.rs
+++ b/api/src/server/auth.rs
@@ -1,20 +1,21 @@
 use crate::{
     config::{OpenIdConfig, ProviderConfig},
     error::ResponseError,
-    models::{NewUser, NewUserProvider, User, UserProvider},
-    schema::{user_providers, users},
+    models::NewUserProvider,
+    schema::user_providers,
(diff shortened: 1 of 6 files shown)


api/src/server/auth.rs
```diff
@@ -199,50 +152,63 @@ pub async fn route_login(
...
+        })?;
+
+    // Add a cookie which can be used to auth requests. We use the UserProvider
+    // ID so that this works even if the User object hasn't been created yet.
+    identity.remember(user_provider_id.to_string());
+
+    // Redirect to the path specified in the OpenID state param
+    Ok(HttpResponse::Found()
+        .header(http::header::LOCATION, redirect_dest)
+        .finish())
```
LucasPickering, line 211: This is mostly just cleanup. I deleted the `init_user` function since we don't need it anymore, and rearranged the logic here to try to clean it up. I also got rid of the `log_in_user` function since it was only a couple lines and after moving the logic around, it was only being called once from here anyway

PR #112, opened 2020-07-19
Fix frontend-prd build

LucasPickering:
Turns out I forgot a couple steps in the FE build. Only reason it was working is because it was copying in some files from my local build system when i tested locally. I ignored a bunch more files, hopefully it works in the CI now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,7 @@
 .vscode/
 pkg/
-node_modules/
-analysis/
-build/
-target/
+**/node_modules/
+**/build/
+**/target/
 **/__generated__/
 api/config/dev.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,8 +1,7 @@
 .vscode/
 pkg/
-node_modules/
-analysis/
-build/
-target/
+**/node_modules/
+**/build/
+**/target/
```
LucasPickering, line 5: Our `.dockerignore` file is just a link to the `.gitignore`. I thought the two used the same syntax but apparently not because docker was copying in the node_modules and other stuff. This should make it behave as intended.

PR #115, opened 2020-07-20
Make the webserver read its DB URL from the config object instead of from `DATABASE_URL`

LucasPickering:
Diesel's CLI pulls its DB url from the env var `DATABASE_URL`, so we've been reading that for gdlk as well. But now in production we want the CLI and GDLK to use separate users, so this is a good opportunity to read it from the config instead for the API code. So the downside is that now we have to define the DB URL in multiple places in dev, but IMO it's more clear and the security buff in production is great too.

Also, I removed dockerize from the image in favor of just using `?connect_timeout=10`.

I was testing this with my deployment machine so this is actually already running in prod.

### DIFF
--- a/api/docker/cmd.sh
+++ b/api/docker/cmd.sh
@@ -1,6 +1,25 @@
-#!/bin/sh
+#!/bin/bash
 
 set -ex
 
-diesel migration run
-./gdlk_api
+# Env variables we expect as input:
+# - GDLK_DB_HOST
+# - GDLK_DB_NAME
+# - GDLK_ROOT_DB_USER
+# - GDLK_ROOT_DB_PASSWORD_FILE
+# - GDLK_APP_DB_USER
+# - GDLK_APP_DB_PASSWORD_FILE
+# - GDLK_SECRET_KEY_FILE
+# - GDLK_GOOGLE_CLIENT_ID_FILE
+# - GDLK_GOOGLE_CLIENT_SECRET_FILE
+
+# We use the DB superuser for migrations so that we can do extension and table alertations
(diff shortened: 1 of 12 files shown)


api/docker/cmd.sh
```diff
@@ -1,6 +1,25 @@
...
+# We use the DB superuser for migrations so that we can do extension and table alertations
+DATABASE_URL="postgres://${GDLK_ROOT_DB_USER}:$(cat $GDLK_ROOT_DB_PASSWORD_FILE)@${GDLK_DB_HOST}/${GDLK_DB_NAME}?connect_timeout=10" \
+    diesel migration run
+
+# We use an unprivileged user for the app cause security
+GDLK_DATABASE_URL="postgres://${GDLK_APP_DB_USER}:$(cat $GDLK_APP_DB_PASSWORD_FILE)@${GDLK_DB_HOST}/${GDLK_DB_NAME}?connect_timeout=10" \
+GDLK_SECRET_KEY=$(cat $GDLK_SECRET_KEY_FILE) \
+GDLK_OPEN_ID__PROVIDERS__GOOGLE__CLIENT_ID=$(cat $GDLK_GOOGLE_CLIENT_ID_FILE) \
+GDLK_OPEN_ID__PROVIDERS__GOOGLE__CLIENT_SECRET=$(cat $GDLK_GOOGLE_CLIENT_SECRET_FILE) \
+    ./gdlk_api
```
LucasPickering, line 25: Previously we pulled in a script that would dynamically read all of the docker secrets in `/run/secrets` and add them to the env based on their file name. I think being this explicit is better though because now each command doesn't get any more credentials than it needs, which limits the impact if we have a vulnerability somewhere.

PR #116, opened 2020-07-21
 setup custom ace mode to highlight comments

### LINKED ISSUE #114: Add ctrl + / to ide for adding comments

JRMurr:
Setup a custom ace mode to highlight comments and tell it what our comment char is to allow for `ctrl+\` to do block comments to close #114.

Its more jank then i would like but ace is a lil jank


frontend/src/util/ace_mode.ts
```diff
@@ -0,0 +1,29 @@
+import 'ace-builds/src-noconflict/mode-plain_text';
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
LucasPickering, line 3: Bruh

PR #123, opened 2020-08-13
Add proper permission checking for mutations

LucasPickering:
Added permission checking. This includes some new tables:
- `roles`
- `user_roles`
- `permissions`
- `role_permissions`

So we use roles to group permissions, then assign roles to users to give them permissions.

### DIFF
--- a/api/migrations/2020-07-25-172603_add_permission_models/up.sql
+++ b/api/migrations/2020-07-25-172603_add_permission_models/up.sql
@@ -1,0 +1,42 @@
+CREATE TYPE role_type AS ENUM ('admin', 'spec_creator');
+CREATE TABLE roles (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    name role_type NOT NULL UNIQUE,
+    is_admin BOOLEAN NOT NULL DEFAULT FALSE -- admins implicitly get all permissions
+);
+
+CREATE TABLE user_roles (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    user_id UUID NOT NULL REFERENCES users(id),
+    role_id UUID NOT NULL REFERENCES roles(id),
+    UNIQUE(user_id, role_id)
+);
+
+CREATE TYPE permission_type AS ENUM (
+    'create_specs',
+    'modify_all_specs',
+    'delete_all_specs',
+    'view_all_user_programs'
+);
+CREATE TABLE permissions (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    name permission_type NOT NULL UNIQUE
+);
+
+CREATE TABLE role_permissions (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    role_id UUID NOT NULL REFERENCES roles(id),
+    permission_id UUID NOT NULL REFERENCES permissions(id),
+    UNIQUE(role_id, permission_id)
+);
+
+INSERT INTO roles (name, is_admin) VALUES ('admin', TRUE);
+INSERT INTO roles (name) VALUES ('spec_creator');
+INSERT INTO permissions (name) VALUES
+    ('create_specs'),
+    ('modify_all_specs'),
+    ('delete_all_specs'),
+    ('view_all_user_programs');
+INSERT INTO role_permissions (role_id, permission_id)
+    SELECT roles.id, permissions.id FROM roles, permissions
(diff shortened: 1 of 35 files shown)


api/migrations/2020-07-25-172603_add_permission_models/up.sql
```diff
@@ -0,0 +1,40 @@
...
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    name permission_type NOT NULL UNIQUE
+);
+
+CREATE TABLE role_permissions (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    role_id UUID NOT NULL REFERENCES roles(id),
+    permission_id UUID NOT NULL REFERENCES permissions(id),
+    UNIQUE(role_id, permission_id)
+);
```
LucasPickering, line 30: @JRMurr how do you feel about these being enums? It's nice cause it makes it easy to map to Rust enums and make sure we don't get invalid names used. But idk how migration will be if we want to add/remove options.

api/src/models/user.rs
```diff
@@ -29,6 +32,28 @@ impl User {
     ) -> dsl::Filter<users::table, WithUsername<'_>> {
         users::table.filter(Self::with_username(username))
     }
+
+    /// Build a query to grant each of the given roles to the user.
+    pub fn add_roles_x(
```
LucasPickering, line 37: I made this `_x` to indicate that this actually executes the query, since our model methods usually just build a query and return it. I couldn't figure out the diesel types to return the query so i figured we'd just take in the connection here and execute it.

api/migrations/2020-07-25-172603_add_permission_models/up.sql
JRMurr: I like the enum, looks like we could use https://www.postgresql.org/docs/9.5/sql-altertype.html when we migrate

api/src/models/user.rs
JRMurr: I might be dumb but is this not async?

api/src/schema.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/api/src/schema.rs b/api/src/schema.rs
+index d7caa53..abdac0c 100644
+--- a/api/src/schema.rs
++++ b/api/src/schema.rs
+@@ -1,6 +1,8 @@
++#![allow(unused_imports)]
++
```
JRMurr, line 7: is this stuff supposed to be here?

api/src/views/context.rs
```diff
@@ -79,13 +91,53 @@ impl UserContext {
...
+                    .inner_join(
+                        user_roles::table.left_join(
+                            // LEFT JOIN here so that even if the role has no
+                            // permissions, we still get its is_admin value
+                            roles::table.left_join(
+                                role_permissions::table
+                                    .left_join(permissions::table),
+                            ),
+                        ),
+                    )
```
JRMurr, line 107: these joins lol 

giphy

api/src/views/context.rs
```diff
@@ -79,13 +91,53 @@ impl UserContext {
...
+                    .select((
+                        permissions::columns::name.nullable(),
+                        roles::columns::is_admin,
+                    ))
+                    .distinct()
+                    .filter(users::columns::id.eq(user_id))
+                    .get_results::<(Option<PermissionType>, bool)>(conn)?;
```
JRMurr, line 114: Im not sure if it would be easy in disel but doing an array agg on the permissions might let use remove the code block after this that makes the hash set. If its annoying im cool with leaving it as is

api/src/views/hardware_spec.rs
```diff
@@ -60,7 +67,16 @@ pub struct UpdateHardwareSpecView<'a> {
...
 
-    fn execute(&self) -> ResponseResult<Self::Output> {
+    fn check_permissions(&self) -> ResponseResult<()> {
+        let user = self.context.user()?;
+        if user.has_permission(PermissionType::ModifyAllSpecs) {
+            Ok(())
+        } else {
+            Err(ClientError::PermissionDenied.into())
+        }
+    }
```
JRMurr, line 77: Might not be worth it since its so simple but could we abstract this func a bit? Seems the same thing is used in a few of the views but with a different permission type.

LucasPickering:
@JRMurr I'm making a ton of changes on this still, and I think that this PR is just too big. I'm gonna split it into 2 PRs, one for just the DB changes and then one for the permission stuff on top.

PR #131, opened 2020-09-02
Add models and DB tables for permissions

LucasPickering:
This is half of the previous PR, just the DB stuff and the diesel models. I scrapped the custom pg enum because I think it will be too annoying to maintain (apparently removing values from an enum is really annoying, which we may want to do). This also prevents having to do some annoying diesel stuff to Rust types for the custom enums, which I think will make it all easier to understand. There is a little bit of manual conversion we have to do now between strings and our `PermissionType` enum, but I don't think it's going to be any problem and I included tests to make sure the enums always stay up to date.

### DIFF
--- a/api/migrations/2020-07-25-172603_add_permission_models/up.sql
+++ b/api/migrations/2020-07-25-172603_add_permission_models/up.sql
@@ -1,0 +1,44 @@
+CREATE TABLE roles (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    slug VARCHAR(50) NOT NULL UNIQUE CHECK (char_length(slug) > 0),
+    name VARCHAR(50) NOT NULL UNIQUE CHECK (char_length(name) > 0),
+    is_admin BOOLEAN NOT NULL DEFAULT FALSE -- admins implicitly get all permissions
+);
+-- autogenerate slug from name
+CREATE TRIGGER "t_roles_insert" BEFORE INSERT ON "roles"
+FOR EACH ROW EXECUTE PROCEDURE set_slug_from_name();
+
+CREATE TABLE user_roles (
+    id UUID PRIMARY KEY DEFAULT uuid_generate_v4(),
+    user_id UUID NOT NULL REFERENCES users(id),
+    role_id UUID NOT NULL REFERENCES roles(id),
+    UNIQUE(user_id, role_id)
+);
+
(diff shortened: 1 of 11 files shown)


api/migrations/2020-07-25-172603_add_permission_models/up.sql
```diff
@@ -0,0 +1,44 @@
...
+    ('Admin', TRUE),
+    ('Spec Creator', FALSE);
+INSERT INTO permissions (name) VALUES
+    ('Create Specs'),
+    ('Modify All Specs'),
+    ('Delete All Specs'),
+    ('View All User Programs');
+INSERT INTO role_permissions (role_id, permission_id)
+    SELECT roles.id, permissions.id FROM roles, permissions
+    WHERE roles.name = 'Spec Creator' and permissions.name = 'Create Specs';
```
LucasPickering, line 44: @JRMurr Do you think I'm missing any permissions here? Not all of these will be implemented, but I wanted to include everything that I think we'll use at some point.

api/src/error.rs
```diff
@@ -55,6 +55,12 @@ pub enum ClientError {
     #[error("Not logged in")]
     Unauthenticated,
 
+    /// Action cannot be performed because the user doesn't have permission to
+    /// do so. This should only be used when the user is already logged in.
+    /// Equivalent of an HTTP 403.
+    #[error("Insufficient permissions to perform this action")]
+    PermissionDenied,
+
```
LucasPickering, line 63: I'll end up using this in my next PR

api/src/models/user.rs
```diff
@@ -29,6 +32,31 @@ impl User {
...
+                        roles::columns::id,
+                    ))
+                    .filter(roles::columns::name.eq_any(role_names)),
+            )
+            .into_columns((
+                user_roles::columns::user_id,
+                user_roles::columns::role_id,
+            ))
+            .execute(conn)
+    }
```
LucasPickering, line 59: Another thing that I'm not using in this PR, but I'll use in the next one for sure.

api/migrations/2020-07-25-172603_add_permission_models/up.sql
JRMurr: I think that covers it, anything else i would think to modify would be infrastructure stuff that we would do manually

JRMurr (approved):

PR #132, opened 2020-09-05
Add basic permission checking to API layer

LucasPickering:
This is the API layer of the permission stuff. This is kind of a shitty implementation, I have some ideas to improve it and make it more generic so we can share more code, but I just wanna get something functional in for now.

### DIFF
--- a/api/src/views/context.rs
+++ b/api/src/views/context.rs
@@ -4,13 +4,16 @@
 use crate::{
     error::{ClientError, ResponseResult},
     models,
-    schema::{user_providers, users},
+    schema::{
+        permissions, role_permissions, roles, user_providers, user_roles, users,
+    },
     util::PooledConnection,
 };
 use diesel::{
     NullableExpressionMethods, OptionalExtension, PgConnection, QueryDsl,
     RunQueryDsl,
 };
+use std::collections::HashSet;
 use uuid::Uuid;
 
 /// Information on the logged-in user.
@@ -18,6 +21,8 @@ use uuid::Uuid;
 pub struct AuthorizedUser {
     pub id: Uuid,
     pub username: String,
+    pub permissions: HashSet<models::PermissionType>,
+    pub is_admin: bool,
 }
 
 impl AuthorizedUser {
@@ -27,6 +32,12 @@ impl AuthorizedUser {
             id: self.id,
             username: self.username,
         }
+    }
+
+    /// Check if the user has the requested permission. Admins implicitly have
+    /// all permissions, so this also checks `self.is_admin`.
+    pub fn has_permission(&self, permission: models::PermissionType) -> bool {
+        self.is_admin || self.permissions.contains(&permission)
     }
 }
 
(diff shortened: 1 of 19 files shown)


api/src/views/context.rs
```diff
@@ -79,13 +91,57 @@ impl UserContext {
...
+                Some(Self {
+                    user_provider_id,
+                    user: Some(AuthorizedUser {
+                        id: user_id,
+                        username,
+                        permissions,
+                        is_admin,
+                    }),
+                })
+            }
```
LucasPickering, line 144: @JRMurr On the last pr i know you asked if we can combine this. It's definitely possible in SQL but right now diesel doesn't support the aggregations we need properly. I'm pretty sure it will work in diesel 2.0 so I left a note to fix that once 2.0 gets release and we upgrade.

api/src/views/mod.rs
```diff
@@ -22,5 +22,12 @@ pub use user_program::*;
...
 
-    fn execute(&self) -> ResponseResult<Self::Output>;
+    fn check_permissions(&self) -> ResponseResult<()>;
+
+    fn execute_internal(&self) -> ResponseResult<Self::Output>;
+
+    fn execute(&self) -> ResponseResult<Self::Output> {
+        self.check_permissions()?;
+        self.execute_internal()
+    }
```
JRMurr, line 32: Is it possible to have two associated types? It seems `check_permissions` always throws the same error so maybe we could set the second associated type to the desired permissions so we don't need to do the boiler plate in each implementation

JRMurr: it would be weird for the few impls that always return Ok(), so maybe we could just define a lambda that does the perm check? Might be overkill though so im cool with not doing any of this

JRMurr (changes requested):

api/src/views/context.rs
```diff
@@ -147,6 +202,7 @@ impl RequestContext {
 mod tests {
     use super::*;
     use crate::{models::Factory, util};
+    use maplit::hashset;
```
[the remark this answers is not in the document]
LucasPickering: ecks dee

api/src/views/mod.rs
LucasPickering: Yes, multiple associated types is possible, but I have a bigger plan for this. In general we'll have two different ways of rejecting based on permissions. One is "you're allowed to know this thing exists, but you can't do whatever you tried", e.g. if you try to modify a hardware spec since those are public but read only except for us.

The other option is that you shouldn't even know something exists, e.g. if you try to delete someone else's user program (since you do that by UUID). In those cases, we want to pretend like the object doesn't even exist, to prevent data leaks.

My long term plan is to convert this views into sort of "view sets" (that's what Django Rest Framework calls them). You'd have one view set per "resource" (basically per DB table), and each view set would define functionality for reading 1 object, reading multiple, and whatever other mutations are possible. So for example, you'd have one view set for user programs, and as part of that you'd define a function that checks if a user is allowed to read a particular user program. So when you go to mutate a user program, you check for read perm first, and if they don't have that, you pretend like it doesn't exist. But if they have read, then you move onto the write perm check, and if that fails you return `PermissionDenied`.

Does that make sense? I haven't fully fleshed out the idea but that's the gist.

LucasPickering: And with that new scheme it should reduce a lot of the boilerplate

JRMurr: that mostly makes sense, tldr would be this is just be the first check before doing more work?

JRMurr (approved):

Issue #133, opened 2020-09-10
Use pg generated columns for `slug` columns


LucasPickering:
All the `slug` columns that we have are just generated from the `name` column at insert. We should use pg generated columns to clean that up. We'll need to make sure there's still an index on the new col, so that filtering can be turbo fast.

JRMurr:
closing since we dont want slugs to update on edit

PR #138, opened 2020-10-05
Switch to generated cols on slugs

JRMurr:
Ez clap #133


JRMurr:
Closing since we dont want slugs to update on edit

PR #144, opened 2020-10-15
Fix up some transactions

LucasPickering:
We have two places in the code that use transactions. Turns out neither were doing what we wanted. On one, I ditched the transaction and just made it fail if the data gets invalid. On the either, I changed the isolation level so that now the transaction does what we really want. In order to do that though, I had to change the tests on that view to not use a transaction (by default all tests are run within a transaction). Do now when we disable transactions, the test code will wipe the DB tables at the end of the test instead (test transaction will be more reliable but the DB wiping is a good fallback). For now though I just had it wipe the two auth tables (`users` and `user_providers`), mostly cause I'm lazy.

### DIFF
--- a/api/src/server/auth.rs
+++ b/api/src/server/auth.rs
@@ -8,8 +8,7 @@ use crate::{
 use actix_identity::Identity;
 use actix_web::{get, http, post, web, HttpResponse};
 use diesel::{
-    Connection, ExpressionMethods, OptionalExtension, PgConnection, QueryDsl,
-    RunQueryDsl,
+    ExpressionMethods, OptionalExtension, PgConnection, QueryDsl, RunQueryDsl,
 };
 use openidconnect::{
     core::{CoreClient, CoreProviderMetadata, CoreResponseType},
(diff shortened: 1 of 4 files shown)


api/src/server/auth.rs
```diff
@@ -143,40 +142,36 @@ pub async fn route_login(
...
-                        .get_result(conn)?;
-                    Ok(existing_id)
-                }
-                Some(inserted_id) => Ok(inserted_id),
-            }
-        })?;
+            .map_err(ResponseError::from_server_error)?,
+            Some(existing_id) => existing_id,
+        }
+    };
```
LucasPickering, line 174: This used to do insert first, then if that failed from a conflict, it selected the existing row. It also had a transaction around it but it turns out that wasn't accomplishing anything.

I changed it to select first, and if the user_provider doesn't exist, then it inserts it. So now the insert can only get a conflict because of a race condition (this endpoint executing twice for the same user simultaneously). And instead of doing nothing on conflict, it now triggers a 500. Techincally it should be a 400 or something, but I don't think it's worth the effort to convert that because it would require a user to log in **for the first time** twice at once, which is exceptionally unlikely.

Also by doing the select first it should make non-first-time logins (which I expect to be the vast majority) faster.

api/src/views/user.rs
```diff
@@ -35,42 +35,49 @@ impl<'a> View for InitializeUserView<'a> {
...
-                    } else {
-                        Ok(created_user)
-                    }
-                })
+                        if updated_rows == 0 {
+                            Err(ClientError::NotFound { source: None }.into())
+                        } else {
+                            Ok(created_user)
+                        }
+                    })
```
LucasPickering, line 80: Basically all I did here was convert the default transaction (which is isolation level `READ COMMITTED`) to the next highest isolation level, `READ REPEATABLE`. We do an insert-then-update here, like so:

1. Insert new user
2. Update one row user_providers to point to that new user

So previously there was a potential race condition if this ran twice simultaneously: Both handlers insert, then the first updates the row to point to its user, then the 2nd does the same update. So you end up with a dangling `users` row from the first request.

With `READ REPEATABLE`, the 2nd request that comes through will update the user_providers row, but when it tries to commit, it will fail because the 1st request modified that row.

api/tests/utils/mod.rs
```diff
@@ -125,3 +143,21 @@ impl QueryRunner {
...
+        // truncate all tables here, but that sounds like a lot of work that I
+        // don't wanna do so just sticking with users for now.
+        let conn = self.db_conn();
+        if (conn.transaction_manager() as &dyn TransactionManager<PgConnection>)
+            .get_transaction_depth()
+            == 0
+        {
+            // TODO clean all tables here
+            diesel::delete(user_providers::table).execute(conn).unwrap();
+            diesel::delete(users::table).execute(conn).unwrap();
```
JRMurr, line 160: This is just me spit balling but could we do some derive macro jank on all disel tables that would add them to a lazy static list or something so we could iter over all of them to delete?

LucasPickering: Yeah I think that would work, we would just have to remember to include that derive on the model when we add a new table. Writing a derive macro is also kind of a bitch. Tbh I think I'd rather just maintain a manual list, but I'd think about it more.

JRMurr: yea it would be ass so just something to keep in mind

JRMurr (approved):

PR #147, opened 2020-10-21
Add execution mutation to GQL schema

LucasPickering:
Adds some minimal API shit to allow executing programs server side. This will execute the program and, if successful, store statistical perf data on it. The API isn't fully fleshed out, but all the data we need is being stored. The next step is to add more fields into the API output that we have (hence all the TODOs). After that we can add some more API stuff to actually read leaderboard data, then after that we can do UI.

### DIFF
--- a/api/schema.graphql
+++ b/api/schema.graphql
@@ -911,7 +911,7 @@ type ProgramRuntimeError {
 Execution output when a program compiles and executes without error, but the
 final state doesn't match the expected state (as defined by the program spec).
 """
-type ProgramFailureOutput {
+type ProgramRejectedOutput {
   # TODO add machine
   """
   TODO
@@ -925,7 +925,7 @@ Execution output when a program executes successfully.
 2. It executed without error
 3. At the end, the input is empty and the output matches the program spec
 """
-type ProgramSuccessOutput {
+type ProgramAcceptedOutput {
   # TODO add machine + stats
   """
   TODO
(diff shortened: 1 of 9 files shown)


api/schema.graphql
```diff
@@ -850,3 +864,96 @@ type DeleteUserProgramPayload {
...
+Execution output when a program compiles and executes without error, but the
+final state doesn't match the expected state (as defined by the program spec).
+"""
+type ProgramFailureOutput {
+  # TODO add machine
+  """
+  TODO
+  """
+  todo: Int! @juniper(infallible: true, ownership: "owned")
+}
```
JRMurr, line 920: maybe `programRejected` or denied?

JRMurr (changes requested):

api/schema.graphql
```diff
@@ -952,7 +952,7 @@ type ExecuteUserProgramPayload {
   outcome, which could be one of:
   - Compile error
   - Runtime error
-  - Failed execution (incorrect input/output at termination)
+  - Rejected execution (incorrect input/output at termination)
   - Success
```
JRMurr, line 956: Really minor but missed accepted here

JRMurr (changes requested):

api/schema.graphql
LucasPickering: Yeah ik I was just gonna leave that one cause I don't think it makes a diff

JRMurr: good enough for me lol

JRMurr (approved):

PR #148, opened 2020-10-21
Flesh out leaderboard API more

LucasPickering:
Adds more output fields to the `executeUserProgram` API. This should be everything we need for that API, after this the next thing we need is fields to read leaderboard stats for a particular program_spec.

You can basically just focus on the API schema and make sure that looks good. All the rust code is pretty straightforward from there. There are a bunch of fields (particularly in `ProgramError` and `MachineState`) that I left out just because we don't need them now. Would be pretty easy to add them later if we need them.

### DIFF
--- a/api/src/util/auth.rs
+++ b/api/src/util/auth.rs
@@ -15,7 +15,6 @@ pub async fn oidc_http_client(
 ) -> Result<openidconnect::HttpResponse, ActixClientError> {
     // There are some unfortunate clones in here because the types we have don't
     // always allow moving the data.
-    // TODO there's probably still room for improvement here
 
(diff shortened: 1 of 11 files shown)


api/src/util/auth.rs
```diff
@@ -68,7 +67,6 @@ pub async fn oidc_request_token(
         Some(id_token) => {
             // TODO better nonce handling here
             match id_token.claims(&token_verifier, &Nonce::new("4".into())) {
-                // TODO remove clone
```
LucasPickering, line 71: I went through all our TODOs and removed a few that we're definitely never gonna fix.

JRMurr (approved):

PR #155, opened 2020-11-01
Auto-save programs and some cleanup

LucasPickering:
- Replaced the Save button on the IDE with an auto-save feature
  - If the user stops editing for 1 second, it saves their changes
- Moved a bunch of logic from `ProgramIde` into a separate hook called `useCompiler`
  - I didn't change the logic at all, but this should make it easier to work on the compiler mgt logic without that component getting huge

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,6 @@ volumes:
   api_build:
   frontend_core_build:
   frontend_wasm_build:
-  frontend_wasm_pkg:
   frontend_node_modules:
 
 services:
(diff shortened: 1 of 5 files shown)


docker-compose.yml
```diff
@@ -46,7 +45,6 @@ services:
       - ./:/app:rw
       - frontend_core_build:/app/target
       - frontend_wasm_build:/app/wasm/target
-      - frontend_wasm_pkg:/app/wasm/pkg
```
LucasPickering, line 49: I removed this so that when the compiler runs inside the container, it updates the type bindings outside. Otherwise, you have to manually run the wasm compiler every time you make changes to update the type bindings.

JRMurr (approved):

PR #161, opened 2020-11-03
Run OIDC provider init concurrently

LucasPickering:
Made provider init concurrent. These will still run on one thread, but since the bottleneck is network I/O, that should still improve perf a ton once we start adding more providers. I ran a test just to make sure this works (I just duplicated the google provider):

```
api_1       | making client: google
api_1       | making client: google2
api_1       | done with: google2
api_1       | done with: google
api_1       | [2020-11-03T01:46:56Z INFO  actix_server::builder] Starting 8 workers
api_1       | [2020-11-03T01:46:56Z INFO  actix_server::builder] Starting "actix-web-service-0.0.0.0:8000" service on 0.0.0.0:8000
```

so it's definitely launching all the HTTP requests before waiting for any to resolve.

### DIFF
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -21,6 +21,7 @@ chrono = "0.4"
 config = {version = "0.10", default-features = false, features = ["json"]}
 diesel = {version = "^1.4.3", default-features = false, features = ["chrono", "postgres", "r2d2", "uuidv07"]}
 env_logger = "0.7"
+futures = "0.3"
 gdlk = {path = "../core"}
 juniper = {version = "0.14.2", default-features = false, features = ["chrono"]}
 # juniper-from-schema = "0.5.2"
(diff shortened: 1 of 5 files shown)


api/Cargo.toml
```diff
@@ -21,6 +21,7 @@ chrono = "0.4"
 config = {version = "0.10", default-features = false, features = ["json"]}
 diesel = {version = "^1.4.3", default-features = false, features = ["chrono", "postgres", "r2d2", "uuidv07"]}
 env_logger = "0.7"
+futures = "0.3"
```
LucasPickering, line 24: This is already in our dep tree, we just didn't have to rely on it directly before.

api/src/server/auth.rs
```diff
@@ -100,6 +107,8 @@ pub async fn route_authorize(
         .authorize_url(
             AuthenticationFlow::<CoreResponseType>::AuthorizationCode,
             move || AuthState::new(next).serialize(),
+            // TODO use a real nonce
+            // https://github.com/LucasPickering/gdlk/issues/159
```
LucasPickering, line 111: Unrelated, but I just create an issue for this so I wanted to tag it

JRMurr (approved):

PR #168, opened 2020-11-08
add delete_dangling_records trigger

JRMurr:
Triggered.


api/migrations/2020-11-08-174607_trigger_users/down.sql
```diff
@@ -0,0 +1,2 @@
+DROP TRIGGER IF EXISTS delete_dangling_records;
+DROP FUNCTION IF EXISTS delete_dangling_program_records;
```
LucasPickering, line 2: I think we shouldn't have the `IF EXISTS` here, cause if this DROP fails that means something about it is broken, so we wanna know. Like hypothetically if you misspelled a function name here we wouldn't know.

Also I don't think the test suite actually runs these `down` migrations, we should prob make that happen at some point.

api/migrations/2020-11-08-174607_trigger_users/up.sql
```diff
@@ -0,0 +1,35 @@
...
+        AND upr.cpu_cycles > pbs.cpu_cycles
+        AND upr.instructions > pbs.instructions
+        AND upr.registers_used > pbs.registers_used
+        AND upr.stacks_used > pbs.stacks_used
+        AND upr.id NOT IN (
+            -- get all referenced rows so they wont be deleted
+            SELECT record_id FROM "user_programs" as up
+            WHERE up.user_id = NEW.user_id AND up.program_spec_id = NEW.program_spec_id
+        )
+        AND upr.id <> NEW.id
```
LucasPickering, line 24: IMO using table aliases makes it a bit harder to read (one more thing to remember), do you mind using the full table names instead?

Also could you add a few more comments just to clarify what each part of the WHERE clause is for?

api/migrations/2020-11-08-174607_trigger_users/up.sql
```diff
@@ -0,0 +1,35 @@
...
+        AND upr.program_spec_id = NEW.program_spec_id
+    )
+    DELETE FROM "user_program_records" WHERE id in (
+        SELECT id
+        FROM "user_program_records" as upr, pbs
+        WHERE upr.user_id = NEW.user_id AND upr.program_spec_id = NEW.program_spec_id
+        AND upr.cpu_cycles > pbs.cpu_cycles
+        AND upr.instructions > pbs.instructions
+        AND upr.registers_used > pbs.registers_used
+        AND upr.stacks_used > pbs.stacks_used
```
LucasPickering, line 18: Do you think we should index these cols? Can you check if adding the indexes lets PG not do a scan? I would think pg is smart enough for that, since the rest of the WHERE clause is only on the PK col

api/migrations/2020-11-08-174607_trigger_users/up.sql
```diff
@@ -0,0 +1,35 @@
+CREATE OR REPLACE FUNCTION delete_dangling_program_records()
+    RETURNS trigger AS
+$$
+BEGIN
+    WITH pbs as (
+        SELECT min(cpu_cycles) as cpu_cycles, min(instructions) as instructions, min(registers_used) as registers_used, min(stacks_used) as stacks_used
+        FROM "user_program_records" as upr
+        WHERE upr.user_id = NEW.user_id
```
LucasPickering, line 8: So is `NEW` defined implicitly for insert triggers? I don't see it declared as an arg but idk much about pg funcs

api/tests/test_user_program_record.rs
```diff
@@ -76,3 +79,67 @@ async fn test_load_pbs_x() {
...
+            .user(&user)
+            .program_spec(&prog_spec)
+            // this should be deleted after the third record since its no longer
+            // has any pbs
+            .cpu_cycles(100)
+            .instructions(100)
+            .registers_used(6)
+            .stacks_used(7)
+            .insert(&conn);
+        // user+prog_spec
```
LucasPickering, line 107: ??

api/tests/test_user_program_record.rs
```diff
@@ -76,3 +79,67 @@ async fn test_load_pbs_x() {
...
+    let runner = QueryRunner::new();
+
+    runner.run_with_conn(|conn| {
+        let user = UserFactory::default().username("user1").insert(&conn);
+        let hw_spec = HardwareSpecFactory::default().name("HW1").insert(&conn);
+        let prog_spec = ProgramSpecFactory::default()
+            .hardware_spec(&hw_spec)
+            .name("Prog1")
+            .insert(&conn);
+
```
LucasPickering, line 96: Can you add two more things here:
1. a 2nd record with at least one stat that matches the PB (e.g. cpu_cycles=1), just to make sure we save _all_ the records that have the min, not just one?
2. A user_program that references one record that has all shit stats, so we know it doesn't try to delete non-PBs that are referenced by `user_programs`

LucasPickering (changes requested):

api/migrations/2020-11-08-174607_trigger_users/up.sql
JRMurr: Yea NEW is always there for insert triggers

api/migrations/2020-11-08-174607_trigger_users/up.sql
JRMurr: Yea ill check if we have them and add them if not

JRMurr: So from the looks of it indexs where you do a `>` or `<` check are not used often. I think if we just put an index on `user_id` `program_spec_id` will be good enough so it would get all records for the current user's program spec then just scan the couple records in there

JRMurr: https://www.postgresql.org/docs/12/indexes-multicolumn.html

api/migrations/2020-11-08-174607_trigger_users/up.sql
```diff
@@ -0,0 +1,58 @@
+CREATE INDEX user_program_records_user_program_ids ON user_program_records (program_spec_id, user_id);
```
LucasPickering, line 1: this name doesn't really seem applicable, the index doesn't include `user_programs`

api/tests/test_user_program_record.rs
```diff
@@ -76,3 +79,103 @@ async fn test_load_pbs_x() {
...
+            .select(user_program_records::columns::id)
+            .get_results::<Uuid>(conn)
+            .unwrap()
+            .into_iter()
+            .collect();
+
+        assert!(!remaining_ids.contains(&to_del.id));
+        assert!(remaining_ids.contains(&to_keep.id));
+        assert!(remaining_ids.contains(&cycle_pb1.id));
+        assert!(remaining_ids.contains(&cycle_pb2.id));
```
LucasPickering, line 179: I think it would be cleaner to just do:

```rust
assert_eq!(
    remaining_ids,
    hashset! { to_del.id, to_keep.id, cycle_pb1.id, cycle_pb2.id }
);
```

(`hashset!` comes from the `maplit` crate)

LucasPickering (changes requested):

LucasPickering (approved):

PR #170, opened 2020-11-12
Use Docker buildkit in CI

LucasPickering:
GitHub's CI supports docker buildkit, which lets us do some neat shit. One thing it supports is localized `.dockerignore` files. For any dockerfile, you can do `<dockerfile name>.dockerignore`. So each build we can specify exactly what files to include. This should limit the build context a ton and make us only build when we need to, instead of rebuilding everything all the time. I went over each image and set it up to only have access to exactly what it needs.

Also I made the CI build (but not push) the prd images on all PRs, instead of just on master.


crates/api/prd.Dockerfile
```diff
@@ -2,7 +2,9 @@
 FROM gcr.io/gdlkit/gdlk-api:latest as rust-builder
 # We need core/, api/, and a bunch of other files, so just copy everything in
 COPY . /app
-RUN cargo build --release
+# TODO remove this
+RUN find /app
+RUN cargo build -p gdlk_api --release
```
JRMurr, line 7: pls remove me

JRMurr (changes requested):

JRMurr (approved):

Issue #166, opened 2020-11-08
Add Composite Score


LucasPickering:
We want a composite score that can summarize all stats of a `user_program`. We should push this as the primary metric. It can be a generated column on the `user_program_records` table.

PR #172, opened 2020-11-13
Add cost stat #166

JRMurr:
We can tweak the sick math on it later

### DIFF
--- a/crates/api/schema.graphql
+++ b/crates/api/schema.graphql
@@ -512,8 +512,7 @@ type UserProgramRecordNode implements Node {
   """
   stacksUsed: Int! @juniper(infallible: true, ownership: "owned")
   """
-  A composite score equal to 2 * cpuCycles + 3 * instructions +
-  100 * registersUsed + 200 * stacksUsed
+  A composite score of the other stats
   """
   cost: Int! @juniper(infallible: true, ownership: "owned")
   """
--- a/crates/api/schema.graphql
+++ b/crates/api/schema.graphql
@@ -512,6 +512,11 @@ type UserProgramRecordNode implements Node {
   """
   stacksUsed: Int! @juniper(infallible: true, ownership: "owned")
   """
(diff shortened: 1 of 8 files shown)


crates/api/schema.graphql
```diff
@@ -512,6 +512,11 @@ type UserProgramRecordNode implements Node {
   """
   stacksUsed: Int! @juniper(infallible: true, ownership: "owned")
   """
+  A composite score equal to 2 * cpuCycles + 3 * instructions +
+  100 * registersUsed + 200 * stacksUsed
```
LucasPickering, line 516: I don't think we should include the formula here, cause 1. it's SECRET SAUCE and 2. we will 100% forget to update it

crates/api/src/models/user_program_record.rs
```diff
@@ -48,6 +48,9 @@ pub struct UserProgramRecord {
     pub stacks_used: i32,
     /// When this row was created, which is also when the program was executed.
     pub created: DateTime<Utc>,
+    /// composite stat, is equal to CEIL(2 * cpu_cycles + 3 * instructions +
+    /// 100 * registers_used + 200 * stacks_used)
```
LucasPickering, line 52: Same here, i think we should just say "see the db schema for actually formula"

crates/api/src/models/user_program_record.rs
```diff
@@ -63,6 +66,9 @@ pub struct UserProgramRecordStats {
     pub registers_used: i32,
     /// Number of unique stacks referenced by the program
     pub stacks_used: i32,
+    /// composite stat, is equal to 2 * cpu_cycles + 3 * instructions +
+    /// 100 * registers_used + 200 * stacks_used
```
LucasPickering, line 70: bro ur killin me

LucasPickering (changes requested):

LucasPickering (approved):

